PR #1, opened 2024-01-22
`pydata/sparse` wrappers for levels, `COO`, and `CSC/CSR` formats

mtsokol:

Here's a PR with a draft implementation of Python wrappers for Tensor constructors, levels, COO, and CSC formats. Some comments:

- `test_coo` and `test_csc` show the construction of fibers using the same code as Hameer's notebook (which adheres to `pydata/sparse`). Tomorrow I will add `CSR` and `DOK/SparseHash` so the whole notebook can be executed (except for `GCXS`).
- Each wrapper has a `_obj` attribute which is the underlying Julia object.
- `Tensor` can be constructed by providing `levels` description (wrapper classes!) and NumPy/Finch array, or just Finch object directly (`jl_data`).
- Only simple element-wise operations are executed here.
- There's also an implementation of the `todense` function that is heavily used in `pydata/sparse`. Here it's done by constructing a new `lvl` description with `Dense` levels only and creating a new Tensor.
- In a separate PR (I think?) I will add CI stages for running those tests.

### Questions:
1. Is it possible to access `fill_value` in `Element`? I mean the first type parameter in `Element{0.0, Float64, Int64}(...)`
2. How can I get the shape tuple of `Tensor`? Right now I obtain shape by iterating through `level`s and aggregating `shape` into one tuple.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,13 +29,13 @@ jobs:
           key:
             test-${{ matrix.os }}-conda-py${{ matrix.python }}-${{ env.CACHE_NUMBER }}-${{
             hashFiles('ci/environment.yml') }}
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: conda-incubator/setup-miniconda@v3
         with:
           activate-environment: finch-tensor-dev
           allow-softlinks: true
           environment-file: ci/environment.yml
           python-version: ${{ matrix.python }}
-          miniforge-variant: Miniforge
+          miniforge-version: latest
           use-only-tar-bz2: true
           use-mamba: true
       - name: Install package
@@ -43,7 +43,7 @@ jobs:
           pip install -e .[tests]
       - name: Run tests
         run: |
-          pytest tests -vv
+          pytest --pyargs finch
       - uses: codecov/codecov-action@v3
         if: always()
       - name: Publish Unit Test Results
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,11 @@ requires-python = ">=3.8"
 dependencies = ["juliapkg>=0.1.10", "juliacall>=0.9.15"]
 
 [project.optional-dependencies]
-tests = ["pytest>=3.5", "sparse>=0.15.1"]
+tests = [
+    "pytest>=3.5",
+    "pytest-cov",
+    "sparse>=0.15.1",
+]
 
 [project.urls]
 Source = "https://github.com/willow-ahrens/finch-tensor"
--- a/finch/__init__.py
+++ b/finch/__init__.py
@@ -1,2 +1,6 @@
-from .tensor import Tensor
-from .tensor import fsprand
+from .tensor import (
+    Tensor, Dense, Element, Pattern, SparseList, SparseByteMap,
+    RepeatRLE, SparseVBL, SparseCOO, SparseHash,
+)
+from .tensor import fsprand, jl
+from .tensor import COO, CSC, CSR
--- a/finch/__init__.py
+++ b/finch/__init__.py
@@ -3,4 +3,4 @@ from .tensor import (
     RepeatRLE, SparseVBL, SparseCOO, SparseHash,
 )
 from .tensor import fsprand, jl
-from .tensor import COO, CSC, CSR
+from .tensor import COO, CSC, CSF, CSR
--- a/finch/tensor.py
+++ b/finch/tensor.py
@@ -1,18 +1,180 @@
+from abc import abstractmethod
+from typing import Tuple
+
+import numpy as np
+import juliacall as jc
+
 from .julia import jl
 
-class Tensor:
-    def __init__(self, data):
-        self.data = data
-    def ndim(self):
-        return jl.ndims(self.data)
-    def dtype(self):
-        return jl.eltype(self.data)
+class _Display:
+    def __repr__(self):
+        return jl.sprint(jl.show, self._obj)
+    def __str__(self):
+        return jl.sprint(jl.show, jl.MIME("text/plain"), self._obj)
+
+
+class Tensor(_Display):
+    def __init__(self, lvl=None, arr=None, jl_data=None):
+        if arr is not None and not isinstance(arr, np.ndarray):
+            raise ValueError("For now only numpy input allowed")
+
+        if lvl is not None and arr is not None and jl_data is None:
+            self._obj = jl.Tensor(lvl._obj, arr)
+        elif jl_data is not None:
+            self._obj = jl_data
+        else:
+            raise ValueError(
+                "either `lvl` and numpy `arr` should be provided or a raw julia object."
+            )
+
+    @property
+    def dtype(self) -> jc.TypeValue:
+        return jl.eltype(self._obj)
+
+    @property
+    def ndim(self) -> int:
+        return jl.ndims(self._obj)
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        return jl.size(self._obj)
+
+    @property
+    def size(self) -> int:
+        return np.prod(self.shape)
+
     def __pos__(self):
-        return Tensor(jl.Base.broadcast(jl.seval("+"), self.data))
-    def __neg__(self):    
-        return Tensor(jl.Base.broadcast(jl.seval("-"), self.data))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj))
+
     def __add__(self, other):
-        return Tensor(jl.Base.broadcast(jl.seval("+"), self.data, other.data))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj, other._obj))
+
+    def __mul__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("*"), self._obj, other._obj))
+
+    def __sub__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("-"), self._obj, other._obj))
+
+    def __truediv__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("/"), self._obj, other._obj))
+
+    def todense(self) -> np.ndarray:
+        shape = jl.size(self._obj)
+
+        # create fully dense array and access `val`
+        dense_lvls = jl.Element(jl.default(self._obj))
+        for _ in shape:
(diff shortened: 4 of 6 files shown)


willow-ahrens:
This looks awesome! I won't have time for a detailed review until tomorrow, but on first glance this looks great!

willow-ahrens:
To answer your questions:
1. You can access the fill value by calling `default(fbr)`.
2. You can get the shape tuple by calling `size(fbr)`.

hameerabbasi (changes requested):
LGTM overall except the dependency on sparse. I'd invert the depencency, and move those bits to `sparse` instead (otherwise we have a circular depencency, and that isn't nice).

mtsokol:
> LGTM overall except the dependency on sparse. I'd invert the dependency, and move those bits to `sparse` instead (otherwise we have a circular dependency, and that isn't nice).

@hameerabbasi `sparse` is declared only as a tests dependency, as in test suites I want to compare it with `sparse`, is it Ok?

FYI, I added an initial setup for CI, but it didn't run, I guess GitHub workflows need to be enabled in repo's settings?
Also, I moved `pyproject.toml` config from `poetry` to the same as `pydata/sparse` (I'm not familiar with poetry here). Could we follow `pydata/sparse` setup for CI and build process?

hameerabbasi:
> `sparse` is declared only as a tests dependency, as in test suites I want to compare it with `sparse`, is it Ok?

TBH, I'm not entirely sure how `pip` will react here. But I guess we can put that in `sparse` too, if it causes issues, or test against `numpy.array_api`.

> FYI, I added an initial setup for CI, but it didn't run, I guess GitHub workflows need to be enabled in repo's settings?
> Also, I moved `pyproject.toml` config from `poetry` to the same as `pydata/sparse` (I'm not familiar with poetry here). Could we follow `pydata/sparse` setup for CI and build process?

That SGTM, but I guess we should ask @willow-ahrens

@willow-ahrens For context, we follow pre-commit rules for linting and formatting. It's quite nice with auto fixes, even in PRs.

pyproject.toml
```diff
@@ -1,20 +1,15 @@
-[tool.poetry]
+[build-system]
+requires = ["setuptools>=64", "setuptools_scm>=8"]
+build-backend = "setuptools.build_meta"
+
+[project]
 name = "finch-tensor"
 version = "0.1.3"
```
hameerabbasi: `setuptools_scm` would make this dynamic and use git tags to manage versioning.

pyproject.toml
```diff
@@ -1,20 +1,15 @@
...
+maintainers = [{ name = "Willow Ahrens", email = "willow.marie.ahrens@gmail.com" }]
 readme = "README.md"
-packages = [{include = "finch", from = "src"}]
-
-[tool.poetry.dependencies]
-python = "^3.8"
-juliapkg = "^0.1.10"
-juliacall = "^0.9.15"
+requires-python = ">=3.8"
+dependencies = ["juliapkg>=0.1.10", "juliacall>=0.9.15", "semantic-version>=2.10.0"]
```
hameerabbasi:
```suggestion
dependencies = ["juliapkg>=0.1.10", "juliacall>=0.9.15"]
```

Already handled by `setuptools_scm` I think.

willow-ahrens:
@mtsokol 
> sparse is declared only as a tests dependency, as in test suites I want to compare it with sparse, is it Ok?

I think that's fine for now! Eventually we may want to use an independent source of truth, but this is fine for now. I don't think circular dependencies will be an issue because it's not finch that depends on sparse, the test environment depends on sparse. The only issue is that if sparse ever begins to use Finch, then testing finch against sparse is just testing finch against itself.

> FYI, I added an initial setup for CI, but it didn't run, I guess GitHub workflows need to be enabled in repo's settings?

Sometimes CI needs to be committed to main before it runs. Maybe make a separate PR for CI and we can merge it, then use this branch to test it?

> Also, I moved pyproject.toml config from poetry to the same as pydata/sparse (I'm not familiar with poetry here). Could we follow pydata/sparse setup for CI and build process?

For CI, absolutely let's get that going. For build process, I'm open to this but I'd like to discuss in a meeting. Poetry is a popular option and it's really quite nice. I encourage you to give it a try! What in particular do you want to fix about the build process? What led to pydata/sparse using setuptools-scm? If we do find it necessary, could we change the build system in a separate PR?

@hameerabbasi

> For context, we follow pre-commit rules for linting and formatting. It's quite nice with auto fixes, even in PRs.

Yes absolutely, let's get that going in a separate PR and merge it so that we can test it out on this one!

willow-ahrens:
Here's the guide I was following to use Poetry: https://www.freecodecamp.org/news/how-to-build-and-publish-python-packages-with-poetry/

src/finch/tensor.py
```diff
@@ -1,18 +1,179 @@
...
-        self.data = data
-    def ndim(self):
-        return jl.ndims(self.data)
-    def dtype(self):
-        return jl.eltype(self.data)
+
+class _Display:
+    def __repr__(self):
+        return self._obj.__repr__()
+
```
willow-ahrens: This should be:
```
class _Display:
    def __repr__(self):
        return jl.sprint(jl.show, self._obj)
    def __str__(self):
        return jl.sprint(jl.show, jl.MIME("text/plain"), self._obj)
```

willow-ahrens (changes requested):
Modulo my suggested change to printing, I approve of the new functionality. Please make changes to CI as a separate PR. Please also make changes to the build system as another separate PR. Thanks for helping out with this!

willow-ahrens:
Once you add CI and pre-commit hooks in a separate PR, I'll make sure they run and that CI must pass before merging.

src/finch/tensor.py
```diff
@@ -1,18 +1,179 @@
...
+    ) -> jc.AnyValue:
+        ...
+
+
+class CSC(_Compressed2D):
+    def get_jl_data(self, shape, lvl, indptr, indices):
+        return jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1])
+
+
+class CSR(_Compressed2D):
```
willow-ahrens: This is not CSR, unfortunately. You should instead create `jl.swizzle(jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1]), 2, 1)`

willow-ahrens (changes requested):
I didn't catch this the first time around, but CSR is not Sparse(Dense(...)). Rather, it is the transposition of CSC. Sparse(Dense(...)) is a weird format that looks like
```
[ 0 0 0 0;
  1 2 3 4;
  0 0 0 0;
  0 0 0 0;
  6 7 8 9]
```

pyproject.toml
mtsokol: Addressed in #2

pyproject.toml
mtsokol: Addressed in #2

src/finch/tensor.py
mtsokol: Ah, right!
I've got a question about `swizzle` - I understand that this operation creates a new Tensor-like object: How should I convert it to an actual Tensor?

```python
from src import finch
import numpy as np

jl = finch.jl  # to just operate on julia directly
A = np.array([[0,0,4], [1,0,0], [2,0,5], [3,0,0]])

A_sparse = jl.Tensor(jl.Dense(jl.SparseList(jl.Element(0.0))), A)
A_sparse + A_sparse  # works

result = jl.swizzle(A_sparse.lvl, 2, 1)
result + result  # it isn't a Tensor
```

willow-ahrens: Hmm. Currently the mode ordering is handled with a separate object, the SwizzledArray. Perhaps the approach here would be to mirror that for now, using a SwizzledArray Wrapper Object. By default, we should probably reverse row to column major in the wrapper as well, so Finch tensors should be row major by default in python.

hameerabbasi:
From the CI:
```
--cov-report --cov-report html --cov-report=xml --cov-report=term --cov sparse --cov-config .coveragerc tests
```

Looks like we need pytest-cov in the test deps.

mtsokol:
> Looks like we need pytest-cov in the test deps.

@hameerabbasi Now tests are running in the CI. One thing - I changed `Miniforge` variant to the previous `Mambaforge` as the former one caused 404 error `image not found` or something similar.

.github/workflows/ci.yml
```diff
@@ -35,15 +35,15 @@ jobs:
           allow-softlinks: true
           environment-file: ci/environment.yml
           python-version: ${{ matrix.python }}
-          miniforge-variant: Miniforge
+          miniforge-variant: Mambaforge
```
hameerabbasi:
```suggestion
          miniforge-version: latest
```

hameerabbasi: Also we should move to `setup-miniconda@v3`

src/finch/tensor.py
mtsokol: I searched Finch.jl codebase and I found that SwizzleArray is converted to Tensor in one of the extensions:

https://github.com/willow-ahrens/Finch.jl/blob/1c7831df99dfcad763e6f9e90c18d9163833b514/ext/SparseArraysExt.jl#L29-L32

I tried to implement `CSR` the same way:
```python
class CSR(_Compressed2D):
    def get_jl_data(self, shape, lvl, indptr, indices):
        a = jl.swizzle(jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1]), 2, 1)
        return jl.Tensor(jl.Dense(jl.SparseList(jl.Element(0.0))), result)
```
but the last line (taken from the extension) fails with:
```
JuliaError: MethodError: no method matching ndims(::Type{DenseLevel{Int64, SparseListLevel{Int64, Vector{Int64}, Vector{Int64}, ElementLevel{0.0, Float64, Int64, Vector{Float64}}}}})
```

mtsokol:

I pushed the latest changes: There are simple COO, CSF, CSR, and CSC wrappers together with Finch Tensor/Levels wrappers.

In terms of row/column major manipulation, instead of reversing dims I did it by reinterpreting the input as column major on entry `np.array(input, order="F")` and back to row major before returning `np.array(result, order="C")`. I added an order flag check to test to verify that `todense()` returns "C"-order numpy array.

willow-ahrens (approved):

willow-ahrens:
Ah, I see. Let's merge it and make a separate PR to support swizzles.

PR #2, opened 2024-01-24
BLD: Move build to setuptools

mtsokol:

This PR updates `pyproject.toml` to match `pydata/sparse` build setup.


finch/_version.py
```diff
@@ -0,0 +1,16 @@
+# file generated by setuptools_scm
```
hameerabbasi: This file shouldn't be in version control and should be added to `.gitignore`.

hameerabbasi (changes requested):
LGTM, just the one minor change.

finch/_version.py
mtsokol: Ah right - done!

hameerabbasi:
I think the readme needs updating, I'd also search for other mentions of Poetry.

mtsokol:
> I think the readme needs updating, I'd also search for other mentions of Poetry.

Sure! I updated `README.md` and `.gitignore` to remove all occurrences of Poetry.

hameerabbasi (approved):

willow-ahrens:
I don't have much of value to add to this. Using setuptools seems really important to y'all, so I'm happy to go along with it, especially since it seems like setuptools might support an opportunity to add a build step. LMK if there's anything I need to do (i.e. adding pypi keys to github secrets, etc.) to get the next version published.

PR #9, opened 2024-01-26
API: Add support for swizzle arrays and input dtypes

mtsokol:

Here's a PR with a draft support for SwizzleArrays, in a summary:
- `permute_dims` is a "transpose" function, and also the first [Array API compatible](https://data-apis.org/array-api/latest/API_specification/generated/array_api.permute_dims.html) here!
- Here's my understanding of a SwizzleArrays support:
  - transposing a Tensor makes internal `_obj` a lazy SwizzleArray which is materialized only when an operation requires it, e.g. `todense()`. This makes it possible to manage SwizzleArrays by transposing multiple times without intermediate materialization.
  - Here, `todense()` function requires materialization.
  - For some specific permutations I noticed that SwizzleArrays don't match with dense counterparts - here's an issue with julia code to reproduce: https://github.com/willow-ahrens/Finch.jl/issues/387. With that explained, `test_permute_dims` should pass.
  - Apart from `swizzle`, is there any other operation that doesn't require materialization and can be stacked here with SwizzleArray?
  - When materializing, I need a "levels" description that doesn't contain data (only composed layers). I implemented it in `_get_lvl_description` method. Could it be done in a simpler way? Can I extract "levels" structure from `Tensor` without levels' data?
- The PR also contains a change that sets the default value with a dtype of a passed data, so `finch.COO(data).todense()` holds the same dtype as `data`.

WDYT?

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -46,33 +46,64 @@ class Tensor(_Display):
         return np.prod(self.shape)
 
     def __pos__(self):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._get_obj()))
 
     def __add__(self, other):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj, other._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._get_obj(), other._get_obj()))
 
     def __mul__(self, other):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("*"), self._obj, other._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("*"), self._get_obj(), other._get_obj()))
 
     def __sub__(self, other):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("-"), self._obj, other._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("-"), self._get_obj(), other._get_obj()))
 
     def __truediv__(self, other):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("/"), self._obj, other._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("/"), self._get_obj(), other._get_obj()))
 
     def todense(self) -> np.ndarray:
-        shape = jl.size(self._obj)
+        obj = self._get_obj()
+        shape = jl.size(obj)
 
         # create fully dense array and access `val`
-        dense_lvls = jl.Element(jl.default(self._obj))
+        dense_lvls = jl.Element(jl.default(obj))
         for _ in shape:
             dense_lvls = jl.Dense(dense_lvls)
-        dense_tensor = jl.Tensor(dense_lvls, self._obj).lvl
+        dense_tensor = jl.Tensor(dense_lvls, obj).lvl
         for _ in shape:
             dense_tensor = dense_tensor.lvl
 
         result = np.array(dense_tensor.val).reshape(shape, order="F")
         return np.array(result, order="C")
+
+    def _get_obj(self) -> jc.AnyValue:
+        return self._get_materialized_obj() if self._is_swizzle() else self._obj
+
+    def _is_swizzle(self) -> bool:
+        return hasattr(self._obj, "body")
+
+    def _get_materialized_obj(self) -> "Tensor":
+        return jl.Tensor(self._get_lvl_description(), self._obj)
+
+    def _get_lvl_description(self) -> jc.AnyValue:
+        # TODO: move to the top of the file
+        lvls_dict = {"Dense": Dense, "Element": Element, "Pattern": Pattern, "SparseList": SparseList}
+
+        obj = self._obj
+        fill_value = jl.default(obj)
+        while hasattr(obj, "body"):
+            obj = obj.body
+        obj = obj.lvl
+
+        lvls = []
+        while not hasattr(obj, "val"):
+            name = str(obj).split("{")[0]
+            lvls.append(lvls_dict[name])
+            obj = obj.lvl
+
+        descr = Element(fill_value)
+        for lvl in lvls[::-1]:
+            descr = lvl(descr)
+        return descr._obj
 
 
 class COO(Tensor):
@@ -80,7 +111,7 @@ class COO(Tensor):
         assert coords.ndim == 2
         ndim = len(shape)
 
-        lvl = jl.Element(fill_value, jl.Vector(data))
+        lvl = jl.Element(data.dtype.type(fill_value), jl.Vector(data))
         ptr = jl.Vector[jl.Int]([1, len(data) + 1])
         tbl = tuple(jl.Vector(coords[i, :] + 1) for i in range(ndim))
 
@@ -95,12 +126,13 @@ class _Compressed2D(Tensor):
         assert len(shape) == 2
 
         data, indices, indptr = arg
+        dtype = data.dtype.type
         data = jl.Vector(data)
         indices = jl.Vector(indices + 1)
         indptr = jl.Vector(indptr + 1)
 
-        lvl = jl.Element(fill_value, data)
-        self._obj = self.get_jl_data(shape, lvl, indptr, indices, fill_value)
+        lvl = jl.Element(dtype(fill_value), data)
+        self._obj = self.get_jl_data(shape, lvl, indptr, indices)
 
     @abstractmethod
     def get_jl_data(
@@ -114,22 +146,21 @@ class _Compressed2D(Tensor):
 
 
 class CSC(_Compressed2D):
-    def get_jl_data(self, shape, lvl, indptr, indices, fill_value):
+    def get_jl_data(self, shape, lvl, indptr, indices):
         return jl.Tensor(
             jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1])
         )
 
 
 class CSR(_Compressed2D):
-    def get_jl_data(self, shape, lvl, indptr, indices, fill_value):
-        swizzled = jl.swizzle(
+    def get_jl_data(self, shape, lvl, indptr, indices):
+        return jl.swizzle(
             jl.Tensor(
                 jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1])
             ),
             2,
             1,
         )
-        return jl.Tensor(jl.Dense(jl.SparseList(jl.Element(fill_value))), swizzled)
 
 
 class CSF(Tensor):
@@ -137,6 +168,7 @@ class CSF(Tensor):
         assert isinstance(arg, tuple) and len(arg) == 3
 
         data, indices_list, indptr_list = arg
+        dtype = data.dtype.type
 
         assert len(indices_list) == len(shape) - 1
         assert len(indptr_list) == len(shape) - 1
@@ -145,7 +177,7 @@ class CSF(Tensor):
         indices_list = [jl.Vector(i + 1) for i in indices_list]
         indptr_list = [jl.Vector(i + 1) for i in indptr_list]
 
-        lvl = jl.Element(fill_value, data)
+        lvl = jl.Element(dtype(fill_value), data)
         for size, indices, indptr in zip(shape[:-1], indices_list, indptr_list):
             lvl = jl.SparseList(lvl, size, indptr, indices)
 
@@ -155,6 +187,11 @@ class CSF(Tensor):
 
 def fsprand(*args):
     return Tensor(jl_data=jl.fsprand(*args))
+
+
+def permute_dims(x: Tensor, axes: tuple[int, ...]):
+    axes = tuple(i + 1 for i in axes)
+    return Tensor(jl_data=jl.swizzle(x._obj, *axes))
 
 
 # LEVELS
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -1,5 +1,5 @@
 from abc import abstractmethod
-from typing import Tuple
+from typing import Union
 
 import numpy as np
 import juliacall as jc
@@ -16,18 +16,76 @@ class _Display:
 
 
 class Tensor(_Display):
-    def __init__(self, lvl=None, arr=None, jl_data=None):
-        if arr is not None and not isinstance(arr, np.ndarray):
-            raise ValueError("For now only numpy input allowed")
+    """
+    A wrapper class for Finch.Tensor and Finch.SwizzleArray.
 
+    Two constructors are supported: `lvl`+`arr`, or `jl_data`.
+
+    Parameters
+    ----------
+    lvl : AbstractLevel, optional
+        Levels description.
+    arr : ndarray, optional
+        NumPy array that should fill `lvl`.
+    jl_data : Finch.SwizzleArray, optional
+        Raw Julia object.
+    order : str or tuple[int, ...], optional
+        Order of the tensor. The order numbers the dimensions from the fastest to slowest.
+        The leaf nodes have mode `0` and the root node has mode `n-1`. If the tensor was square
+        of size `N`, then `N .^ order == strides`. Available options are "C" (row-major),
+        "F" (column-major), or a custom order. Default: row-major.
+
+    Returns
+    -------
+    Tensor
+        Python wrapper for Finch Tensor.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import finch
+    >>> x = finch.Tensor(lvl=finch.Dense(finch.Element(0)), arr=np.arange(3))
+    >>> x.todense()
+    array([0, 1, 2])
+    """
+    row_major = "C"
+    column_major = "F"
+
+    def __init__(self, lvl=None, arr=None, jl_data=None, order=None):
+        if arr is not None and not isinstance(arr, np.ndarray):
+            raise ValueError("For now only numpy input allowed.")
+
+        # constructor for levels description and NumPy array input
         if lvl is not None and arr is not None and jl_data is None:
-            self._obj = jl.Tensor(lvl._obj, np.array(arr, order="F"))
-        elif jl_data is not None:
+            order = self.preprocess_order(order, arr.ndim)
+            inv_order = tuple(i - 1 for i in jl.invperm(order))
+            self._obj = jl.swizzle(jl.Tensor(lvl._obj, arr.transpose(inv_order)), *order)
+        # constructor for a raw julia object
+        elif jl_data is not None and lvl is None and arr is None:
+            if order is not None:
+                raise ValueError("When passing a julia object order can't be provided.")
+            if not jl.isa(jl_data, jl.Finch.SwizzleArray):
+                raise ValueError("`jl_data` must be a SwizzleArray.")
             self._obj = jl_data
         else:
             raise ValueError(
-                "either `lvl` and numpy `arr` should be provided or a raw julia object."
+                "Either `lvl` and numpy `arr` should be provided or a raw julia object."
             )
+
+    def __pos__(self):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj))
+
+    def __add__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj, other._obj))
+
+    def __mul__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("*"), self._obj, other._obj))
+
+    def __sub__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("-"), self._obj, other._obj))
+
+    def __truediv__(self, other):
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("/"), self._obj, other._obj))
 
     @property
     def dtype(self) -> jc.TypeValue:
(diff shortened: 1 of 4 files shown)


src/finch/tensor.py
```diff
@@ -114,29 +146,29 @@ def get_jl_data(
...
-        swizzled = jl.swizzle(
+    def get_jl_data(self, shape, lvl, indptr, indices):
+        return jl.swizzle(
             jl.Tensor(
                 jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1])
             ),
             2,
             1,
         )
-        return jl.Tensor(jl.Dense(jl.SparseList(jl.Element(fill_value))), swizzled)
```
mtsokol, line 132: Here I get rid of an immediate materialization for CSR format

willow-ahrens:
> Here, `todense()`, `+`, `-`, and other elemwise functions require materialization so a `_get_obj()` is called.

An issue here I think is that when you call todense, you create a column-major array, is that right?

> For some specific permutations I noticed that SwizzleArrays don't match with dense counterparts - here's an issue with julia code to reproduce: [Question regarding 3d swizzle array Finch.jl#387](https://github.com/willow-ahrens/Finch.jl/issues/387). With that explained, `test_permute_dims` should pass.

Perhaps related to the above, you can't just reverse the permutation if the permutation is expected to be applied to a reversed input. We may need to be more precise about which arrays are row or column major and how we convert between.
```
julia> list = [1, 2, 3]
3-element Vector{Int64}:
 1
 2
 3

julia> perm = [3, 1, 2]
3-element Vector{Int64}:
 3
 1
 2

julia> list[perm]
3-element Vector{Int64}:
 3
 1
 2

julia> reverse(list)[reverse(perm)]
3-element Vector{Int64}:
 2
 3
 1

julia> reverse(list)[invperm(reverse(invperm(perm)))]
3-element Vector{Int64}:
 3
 1
 2
```
 
> Apart from `swizzle`, is there any other operation that doesn't require materialization and can be stacked here with SwizzleArray?

There are, but I wouldn't worry about it for now. There are array wrappers (like padding, circular shifting, etc), but I think we should skip that for the time being.

> When materializing, I need a "levels" description that doesn't contain data (only composed layers). I implemented it in `_get_lvl_description` method. Could it be done in a simpler way? Can I extract "levels" structure from `Tensor` without levels' data?

On this point, I think we shouldn't define either `_get_lvl_description` or `_get_materialized_obj`. The swizzle should be treated as a finished tensor format by itself, and if we want to materialize it to a different format or storage order, we can call "copyto".

willow-ahrens:
I'll need a little more time to review this one, because your question about whether the `Tensor` object should wrap every array Julia produces or just Finch.Tensor is a good question.

willow-ahrens:
Julia is 1-indexed, col-major array
Finch is 1-indexed, col-major array
Python is 0-indexed, row-major array
JuliaCall(JuliaArray) is 0-indexed, col-major array

I think the purpose of finch-tensor (this package) should be to define a tensor which, by default, is a 0-indexed, row-major array. Under the hood of Tensor, we should be allowed to do anything we wish to support that. I see a few paths:
0. This PR. It seems like this PR is going in the direction that Tensor just wraps any Julia Array, and we can arbitrarily construct lazy Julia array objects inside of it. This has the benefit that we can really go crazy with what kind of things we put in the Tensor type but I think it would feel safer if the jl object we were wrapping had a more predictable type.
1. The stored jlobj inside of the Tensor is always a column-major Finch.Tensor that we are reinterpreting as row-major, and we do things like `shape(self) = np.reverse(jl.size(self._obj))`. This has the added benefit that we can always expect _obj to be Finch.Tensor, and access it accordingly. permute_dims would be eager unless we use the lazy api.
2. The stored jlobj inside of the Tensor is always a Swizzle, and the user is allowed to specify a mode ordering, which we interpret in a row-major context. Thus, if the user specifies no dimension ordering, the object inside the tensor is `swizzle(Tensor, reverse(1:ndims))`. If the user requests column major, we still store `swizzle(Tensor, 1:ndims)`. The only lazy transformation the user can do outside of the lazy api is transposition.
3. 1, but we handle our own dimension permutation with our own vector of dimensions in Tensor, and construct swizzles as needed to make things work.

I think that because people want to talk about CSC and CSR format, and probably want to use the same `Tensor` type to do it, we should go with 2 or 3. As a development decision, I like the idea of constraining what `Tensor` can wrap. What do others think?

src/finch/tensor.py
```diff
@@ -46,41 +46,72 @@ def size(self) -> int:
...
         result = np.array(dense_tensor.val).reshape(shape, order="F")
         return np.array(result, order="C")
 
+    def _get_obj(self) -> jc.AnyValue:
+        return self._get_materialized_obj() if self._is_swizzle() else self._obj
+
+    def _is_swizzle(self) -> bool:
+        return hasattr(self._obj, "body")
+
+    def _get_materialized_obj(self) -> "Tensor":
```
willow-ahrens: I don't think this should be an option, I think we should view `swizzle(::Finch.Tensor)` as a materialized object and if the user wants to change the order they can copy into a different `Tensor` Object.

willow-ahrens: That said, you can achieve this with `jl.copy(self._obj)`.

src/finch/tensor.py
```diff
@@ -46,41 +46,72 @@ def size(self) -> int:
...
             dense_tensor = dense_tensor.lvl
 
         result = np.array(dense_tensor.val).reshape(shape, order="F")
         return np.array(result, order="C")
 
+    def _get_obj(self) -> jc.AnyValue:
+        return self._get_materialized_obj() if self._is_swizzle() else self._obj
+
+    def _is_swizzle(self) -> bool:
+        return hasattr(self._obj, "body")
```
willow-ahrens: Can't we call jl.typeof here?

willow-ahrens: not that I'm sure we're taking this approach

willow-ahrens:
side note: something I've been thinking about: broadcast dimension padding is reversed in Julia and in Python. you can get correct semantics if you transpose everything first to reverse dims, then transpose back.

mtsokol:
> > Here, `todense()`, `+`, `-`, and other elemwise functions require materialization so a `_get_obj()` is called.
> 
> An issue here I think is that when you call todense, you create a column-major array, is that right?

@willow-ahrens Sorry for lack of context: When I try to e.g. add two SwizzleArrays, it says that there's no matching method.

```julia
sw = swizzle(A_fbr, 2, 1, 3)

sw + sw
# ERROR: MethodError: no method matching +(::Finch.SwizzleArray{(2, 1, 3), Tensor{DenseLevel{…}}}, ::Finch.SwizzleArray{(2, 1, 3), Tensor{DenseLevel{…}}})

A_fbr + sw
# ERROR: MethodError: no method matching +(::Tensor{DenseLevel{Int64, SparseListLevel{…}}}, ::Finch.SwizzleArray{(2, 1, 3), Tensor{DenseLevel{…}}})
```

Therefore I understand that all elementwise functions require materialized (Tensors-only) input, is that correct?

> On this point, I think we shouldn't define either _get_lvl_description or _get_materialized_obj. The swizzle should be treated as a finished tensor format by itself, and if we want to materialize it to a different format or storage order, we can call "copyto".

It's fine for me - then I only need to be able to call all Tensor functions on SwizzleArrays.

hameerabbasi:
> > > Here, `todense()`, `+`, `-`, and other elemwise functions require materialization so a `_get_obj()` is called.
> > 
> > 
> > An issue here I think is that when you call todense, you create a column-major array, is that right?
> 
> @willow-ahrens Sorry for lack of context: When I try to e.g. add two SwizzleArrays, it says that there's no matching method.
> 
> ```julia
> sw = swizzle(A_fbr, 2, 1, 3)
> 
> sw + sw
> # ERROR: MethodError: no method matching +(::Finch.SwizzleArray{(2, 1, 3), Finch.SwizzleArray{…}}, ::Finch.SwizzleArray{(2, 1, 3), Finch.SwizzleArray{…}})
> 
> A_fbr + sw
> # ERROR: MethodError: no method matching +(::Tensor{DenseLevel{Int64, SparseListLevel{…}}}, ::Finch.SwizzleArray{(2, 1, 3), Finch.SwizzleArray{(2, 1, 3), Tensor{…}}})
> ```
> 
> Therefore I understand that all elementwise functions require materialized (Tensors-only) input, is that correct?

It seems to me we have multiple levels of nested `SwizzleArray`s, which Finch doesn't have a rewrite rule for. I guess extracting the axes and internal `Tensor` and then manipulating those so that there is just one level of swizzling would help.

mtsokol:
@hameerabbasi Sorry! I copied output from the wrong line - the same happens for single SwizzleArrays (I updated my comment)

mtsokol:
> What do others think?

@willow-ahrens I like option 3. so everything internally is a column-major `Tensor`, we reinterpret it to NumPy realm when needed, `permute_dims` is eager where permutation order is adjusted to map row-column. 

Then for the sake of the eager mode `permute_dims` won't yield a SwizzleArray but rather `copyto!` it.

willow-ahrens:
I think we should probably meet to discuss the approach and some of the bugs you've been hitting, but overall I'm happiest with 3.

To clarify, ideally we should be able to just add swizzle arrays, but it looks like I didn't implement that (it was a little tricky to figure out which tensors needed transposition). Some of the high-level functions of Finch are not as well tested or supported as the direct finch interface. You are the first developer to really use the high-level finch interface, so if something doesn't feel right it might not be intentional.

As I continue implementing the lazy @compiled interface, I'm realizing that several cases (such as + of two swizzles) might be better expressed using @compiled, so I think using the same code path for this stuff will reduce attack area and make it easier to support more cases.

willow-ahrens:
Regarding permutedims in approach 3: What does numpy do in this situation? Numpy arrays have variable strides, so in theory permutedims could just change the strides on the array. When does Numpy actually transpose and when does it just mess with the strides?

willow-ahrens:
From the meeting, it sounds like we're going with Option 2. I'll try to stay on top of any related Finch issues, thanks!

src/finch/tensor.py
mtsokol: I decided to write this check with:
```python
def _is_swizzle(self) -> bool:
    return jl.isa(self._obj, jl.Finch.SwizzleArray)
```

mtsokol:
> From the meeting, it sounds like we're going with Option 2. I'll try to stay on top of any related Finch issues, thanks!

@willow-ahrens @hameerabbasi I followed option 2 and yesterday's discussion: Right now `_obj` is always a `SwizzleArray` with a default permutation depending on `order` (n...1 when row-major (default) is set or 1...n when column-major is).

When constructing a Tensor a permutation is applied as a `SwizzleArray`. When `todense()` is called to get a numpy array, a materialization occurs (and only then!) and a transposition happens, if row-major is expected.

Nested `SwizzleArrays` aren't allowed. When `permute_dims` is called, the permutation inside `SwizzleArray` is adjusted and returned as `SwizzleArray(Tensor(...))`.

Please take a look at `CSC` and `CSR` classes - now they differ only in `swizzle` permutation argument.

I will add some tests to also check if permutation is correct internally after different operations. Once #390 is fixed the test suite should be green.

Just to repeat from yesterday, I commented out `+ / *` elem-ops on `Tensor` as we need `SwizzleArray` to support it.

src/finch/tensor.py
mtsokol: Sure! I removed `_get_materialized_obj`.

mtsokol:

I finished `SwizzleArray` support here, with the option 2 (my previous comment explains some details). Tensors can be constructed (with different formats), transposed, and converted to dense NumPy.

Maybe we could try to merge it? Would you review it?

src/finch/tensor.py
```diff
@@ -16,17 +15,36 @@ def __str__(self):
...
+    """
+    ROW_MAJOR = "C"
+    COLUMN_MAJOR = "F"
+
+    def __init__(self, lvl=None, arr=None, jl_data=None, order=None):
         if arr is not None and not isinstance(arr, np.ndarray):
-            raise ValueError("For now only numpy input allowed")
+            raise ValueError("For now only numpy input allowed.")
+        if order not in (None, self.ROW_MAJOR, self.COLUMN_MAJOR):
+            raise ValueError(f"order must be a 'C' or 'F', but is: {type(order)}.")
```
willow-ahrens: Let's support arbitrary orders too as an array of dims

src/finch/tensor.py
```diff
@@ -38,105 +56,123 @@ def ndim(self) -> int:
...
     def size(self) -> int:
         return np.prod(self.shape)
 
+    @property
+    def _permutation(self) -> tuple[int, ...]:
+        return jl.typeof(self._obj).parameters[1]
+
+    def _apply_new_permutation(self, value: tuple[int, ...]) -> None:
+        p = self._permutation
+        if self.order == self.ROW_MAJOR:
```
willow-ahrens: could we use a utility function that maps e.g. Row_major to [1, 2, 3, ... n] and then handle the general case in this function and others? We could use the function to parse orders any time we accept them.

src/finch/tensor.py
```diff
@@ -16,17 +15,36 @@ def __str__(self):
 
 
 class Tensor(_Display):
-    def __init__(self, lvl=None, arr=None, jl_data=None):
+    """
+    A wrapper class for Finch.Tensor and Finch.SwizzleArray
+    """
+    ROW_MAJOR = "C"
```
willow-ahrens: can we also do lowercase?

src/finch/tensor.py
```diff
@@ -38,105 +56,123 @@ def ndim(self) -> int:
...
+        dense_tensor = jl.Tensor(dense_lvls, obj).lvl
         for _ in shape:
             dense_tensor = dense_tensor.lvl
 
-        result = np.array(dense_tensor.val).reshape(shape, order="F")
-        return np.array(result, order="C")
+        result = np.array(jl.reshape(dense_tensor.val, shape))
+        return result.transpose(None) if self.order == self.ROW_MAJOR else result
+
+    def _is_swizzle(self) -> bool:
```
willow-ahrens: shouldn't this always be true?

willow-ahrens (changes requested):
This looks really good! I'm pushing for supporting arbitrary orders because np arrays allow arbitrary strides. So we could accept "c", "f", or an index order as input. 

But overall, to be clear, this looks great!

src/finch/tensor.py
```diff
@@ -16,17 +15,36 @@ def __str__(self):
...
+        # constructor for levels description and NumPy array input
         if lvl is not None and arr is not None and jl_data is None:
-            self._obj = jl.Tensor(lvl._obj, np.array(arr, order="F"))
-        elif jl_data is not None:
+            if order == self.ROW_MAJOR or order is None:
+                permutation = tuple(range(1, arr.ndim + 1)[::-1])
+                self.order = self.ROW_MAJOR
+            elif order == self.COLUMN_MAJOR:
+                permutation = tuple(range(1, arr.ndim + 1))
+                self.order = order
```
willow-ahrens: why do we maintain this in the python Tensor class as well as the julia swizzle class? I think it would be easier to maintain one source of truth for the order.

willow-ahrens (changes requested):
Another review just to say that I think we should remove the self.order attribute in favor of a getter method for the order of the underlying swizzle.

src/finch/tensor.py
hameerabbasi: If we want to support arbitrary orders, we can by manually calling `.transpose(axes=...)` on the constructed object; does it make sense to put that into the constructor itself?

mtsokol: FYI I just pushed changes that support arbitrary orders. I will explain that in a separate comment.

src/finch/tensor.py
mtsokol: Right now it's always true - I removed this function!

src/finch/tensor.py

src/finch/tensor.py
mtsokol: Done! I added `def preprocess_order(...)` that maps string names, NoneType and tuples to 1-indexed permutations.

src/finch/tensor.py
mtsokol: I think that we need to maintain both the `order` attribute in the python Tensor and julia swizzle class.

`SwizzleArray` defines **current** permutation, where `order` in the python Tensor **original** permutation.

Let's assume a user creates a tensor that is row-major, so the permutation is `[n, n-1, ..., 1]`. This permutation is set in the `order` attribute and `SwizzleArray` internal object. Then the user performs and arbitrary permutation `p`. This permutation is set to the internal `SwizzleArray`.
Then the user calls `todense()`. We need to know how to move to the NumPy's row-major order. We can't just use what is in `SwizzleArray` as it has user's permutation. We use original `order` to know how to move to the actual row-major after materializing the internal `SwizzleArray`.

1. when created: order: (3, 2, 1), swizzle array: (3, 2, 1)
2. custom user permutation: order (3, 2, 1), swizzle array: (2, 1, 3)
3. calling `todense()`: order (3, 2, 1) <-- we still need it to move back to NumPy!, swizzle array (2, 1, 3) <-- just used for materialization.

src/finch/tensor.py
mtsokol: Right now any order that is a permutation can be passed (and also string values "F" and "C"). `order` remembers that permutation and once we want to move back to numpy with `todense()` it's just a `.transpose(self.get_inv_order())` to move back to numpy order.

src/finch/tensor.py
willow-ahrens: Ah, I see what you're going for here. The idea is that the "original" order of the tensor is what the user would want when they call `todense`. I disagree on this point though, because the user could also create a dense finch tensor and I think we want to have the property that todense() is non-copying whenever possible. I think we want to build a system where the following is always true, right? (ignoring syntax errors and the fact that we don't define a `strides` property for finch tensors):

```
tns = Tensor(Dense(Dense(Dense(Element(0.0))), order=order)
tns_2 = permutedims(tns, some_other_order)
tns_2.strides == tns_2.todense().strides
```

mtsokol: `todense()` is a specific case where it matters to retain "original" `order`. Let me give another example. We have a `Tensor(_obj=SwizzleArray(data, (3, 2, 1)))` and a user wants to transpose dims with permutation: `(1, 3, 2)`. How do we know which dimensions to permute?

If we update `SwizzleArray` directly to `...data, (3, 1, 2)` then it will result with incorrect permutation. 

Here's one more example:

```python
tns = Tensor(np.ones((10, 1, 100)), order=(3, 2, 1))

tns_2 = finch.permute_dims(tns, (1, 3, 2)) # results in SwizzleArray(data, (2, 3, 1)), shape is (10, 100, 1)
```
Then a user wants to permute again (switch second dim with the last one again):
```python
tns_3 = finch.permute_dims(tns_2, (1, 3, 2))
```
Inside we have `SwizzleArray(data, (2, 3, 1))`, how do we know what the resulting order is?
Should we switch 3 with 1? That's incorrect.
**We need to store the original order** of (3, 2, 1), because then we can first inverse what is in the swizzle with the inverse of the original order to: (1, 3, 2), apply desired permutation to get: (1, 2, 3) and then reapply the original order: (3, 2, 1).

(3, 2, 1) order and again a shape of (10, 1, 100) is what user expects:
```python
tns = Tensor(np.ones((10, 1, 100)), order=(3, 2, 1))
tns = finch.permute_dims(tns, (1, 3, 2))
tns = finch.permute_dims(tns, (1, 3, 2))
tns.shape  # (10, 1, 100)
```

willow-ahrens:
A few notes from our discussion:
1. `shape(Tensor(arr, order=order)) == shape(arr)`
2. `shape(permutedims(Tensor(arr, order=order), order2)) == shape(permute_dims(arr, order2))`
3. `to_dense()` should never materialize a dense finch tensor
4. `swizzle(swizzle(arr, order1), order2)` should collapse swizzles
5. the python `Tensor` wrapper only needs to store it's jl obj, which is always `jl.swizzle(jl.Tensor(arr), order)`

hameerabbasi:
> to_dense() should never materialize a dense finch tensor

I'm curious of the reasoning behind this, is there a concrete reason at all?

willow-ahrens:
Both `np.Array` and `finch.Tensor` can represent lazy permutations and any combination of row/column major layout. When `finch.Tensor` is dense, then it seems to make sense that the correspondence between the two representations would be realized in `to_dense()`. While `finch.Tensor` uses a swizzle to manage dimension ordering, `np.Array` uses strides, but one would hope that `np.Array(finch.Tensor)` and `finch.Tensor.to_dense()` could both avoid materialization and do the direct conversion.

src/finch/tensor.py
```diff
@@ -16,17 +16,53 @@ def __str__(self):
...
         if lvl is not None and arr is not None and jl_data is None:
-            self._obj = jl.Tensor(lvl._obj, np.array(arr, order="F"))
-        elif jl_data is not None:
+            order = self.preprocess_order(order, arr.ndim)
+            inv_order = tuple(i - 1 for i in jl.invperm(order))
+            self._obj = jl.swizzle(jl.Tensor(lvl._obj, arr.transpose(inv_order)), *order)
+        # constructor for a raw julia object
+        elif jl_data is not None and lvl is None and arr is None:
+            if order is not None:
+                raise ValueError("When passing a julia object order can't be provided.")
```
willow-ahrens, line 66: maybe also check that the object is a swizzle of tensor?

src/finch/tensor.py
```diff
@@ -38,105 +74,138 @@ def ndim(self) -> int:
...
+
+    def get_inv_order(self, zero_indexing=True) -> tuple[int, ...]:
+        inv_order = jl.invperm(self._order)
+        if zero_indexing:
+            inv_order = tuple(i - 1 for i in inv_order)
+        return inv_order
+
     def __pos__(self):
-        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj))
+        return Tensor(jl_data=jl.Base.broadcast(jl.seval("+"), self._obj), order=self.order)
```
willow-ahrens: is self.order defined?

willow-ahrens: Moreover, does this rely on the broadcast returning a tensor and not just a swizzle?

src/finch/tensor.py
```diff
@@ -16,17 +16,53 @@ def __str__(self):
...
+                permutation = order
+            else:
+                raise ValueError(f"Custom order is not a permutation: {order}.")
+        else:
+            raise ValueError(f"order must be 'C', 'F' or a tuple, but is: {type(order)}.")
+
+        return permutation
+
+
+    def __init__(self, lvl=None, arr=None, jl_data=None, order=None):
```
willow-ahrens, line 54: idk where the docs go, but I would say that order numbers the dimensions from fastest to slowest, so the leaf nodes have mode 0 and the root node has mode n-1. If the tensor were square of size `N`, then `N .^ order == strides`

src/finch/tensor.py
```diff
@@ -38,105 +74,138 @@ def ndim(self) -> int:
...
-
-        # create fully dense array and access `val`
-        dense_lvls = jl.Element(jl.default(self._obj))
-        for _ in shape:
-            dense_lvls = jl.Dense(dense_lvls)
-        dense_tensor = jl.Tensor(dense_lvls, self._obj).lvl
-        for _ in shape:
+        obj = self._obj
+
+        if self._is_dense:
```
willow-ahrens, line 157: This looks great now!

willow-ahrens (commented):
Looks good! A few minor things now related to docs and future behavior about the output of broadcast. (I think we probably shouldn't be re-swizzling the results of broadcast).

src/finch/tensor.py
mtsokol: Done! I added another check.

src/finch/tensor.py
mtsokol: Ah, I forgot to remove `self.order`, done!

So these element-wise operations aren't tested right now as `broadcast` doesn't accept `SwizzleArrays`. I think we discussed it before and it should be testable once we have lazy-mode that can execute these operations with `SwizzleArrays`.

src/finch/tensor.py
mtsokol: I added a docstring that explains it.

mtsokol:
@willow-ahrens I applied all review comments - thank you for feedback!

About zero-copy constructors: let me implement it in a follow-up PR - it will contain zero-copy fully-dense Tensor constructor, and `from_scipy` zero-copy constructors with tests.

willow-ahrens (approved):

hameerabbasi:
Thanks a lot for your work on this, @mtsokol!

PR #10, opened 2024-02-06
API: Add fully-dense no-copy and `scipy/numpy` input `Tensor` constructors

mtsokol:

This PR contains:
1. `finch.Tensor(scipy.sparse.spmatrix)` and `finch.Tensor(numpy.ndarray)` constructors.
2. Support for no-copy constructor for fully dense Tensors.
3. `levels.py` and `utils.py` files to clean up the code a bit. 
4. `Storage` class for device support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ pytest = "^7.4.4"
 pre-commit = "^3.6.0"
 pytest-cov = "^4.1.0"
 sparse = "^0.15.1"
+scipy = "^1.7"
 
 [build-system]
 requires = ["poetry-core"]
--- a/src/finch/__init__.py
+++ b/src/finch/__init__.py
@@ -1,5 +1,4 @@
-from .tensor import (
-    Tensor,
+from .levels import (
     Dense,
     Element,
     Pattern,
@@ -9,9 +8,14 @@ from .tensor import (
     SparseVBL,
     SparseCOO,
     SparseHash,
+    Storage,
+    DenseStorage,
 )
-from .tensor import fsprand, permute_dims
-from .tensor import COO, CSC, CSF, CSR
+from .tensor import (
+    Tensor,
+    fsprand,
+    permute_dims,
+)
 
 __all__ = [
     "Tensor",
@@ -24,10 +28,8 @@ __all__ = [
     "SparseVBL",
     "SparseCOO",
     "SparseHash",
+    "Storage",
+    "DenseStorage",
     "fsprand",
     "permute_dims",
-    "COO",
-    "CSC",
-    "CSF",
-    "CSR",
 ]
--- a/src/finch/levels.py
+++ b/src/finch/levels.py
@@ -1,0 +1,93 @@
+import numpy as np
+
+from .julia import jl
+from .typing import OrderType
+
+
+class _Display:
+    def __repr__(self):
+        return jl.sprint(jl.show, self._obj)
+
+    def __str__(self):
+        return jl.sprint(jl.show, jl.MIME("text/plain"), self._obj)
+
+
+# LEVEL
+
+class AbstractLevel(_Display):
+    pass
+
+
+# core levels
+
+class Dense(AbstractLevel):
+    def __init__(self, lvl, shape=None):
+        args = [lvl._obj]
+        if shape is not None:
+            args.append(shape)
+        self._obj = jl.Dense(*args)
+
+
+class Element(AbstractLevel):
+    def __init__(self, fill_value, data=None):
+        args = [fill_value]
+        if data is not None:
+            args.append(data)
+        self._obj = jl.Element(*args)
+
+
+class Pattern(AbstractLevel):
+    def __init__(self):
+        self._obj = jl.Pattern()
+
+
+# advanced levels
+
+class SparseList(AbstractLevel):
+    def __init__(self, lvl):
+        self._obj = jl.SparseList(lvl._obj)
+
+
+class SparseByteMap(AbstractLevel):
+    def __init__(self, lvl):
+        self._obj = jl.SparseByteMap(lvl._obj)
+
+
+class RepeatRLE(AbstractLevel):
+    def __init__(self, lvl):
+        self._obj = jl.RepeatRLE(lvl._obj)
+
+
+class SparseVBL(AbstractLevel):
+    def __init__(self, lvl):
+        self._obj = jl.SparseVBL(lvl._obj)
+
+
+class SparseCOO(AbstractLevel):
+    def __init__(self, ndim, lvl):
+        self._obj = jl.SparseCOO[ndim](lvl._obj)
+
+
+class SparseHash(AbstractLevel):
+    def __init__(self, ndim, lvl):
+        self._obj = jl.SparseHash[ndim](lvl._obj)
+
+
+# STORAGE
+
+class Storage:
+    def __init__(self, levels_descr: AbstractLevel, order: OrderType = None):
+        self.levels_descr = levels_descr
+        self.order = order if order is not None else "C"
+
+    def __str__(self) -> str:
+        return f"Storage(lvl={str(self.levels_descr)}, order={self.order})"
+
+
+class DenseStorage(Storage):
+    def __init__(self, ndim: int, dtype: np.dtype, order: OrderType = None):
+        lvl = Element(np.int_(0).astype(dtype))
+        for _ in range(ndim):
+            lvl = Dense(lvl)
+
+        super().__init__(levels_descr=lvl, order=order)
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -19,14 +19,16 @@ class Tensor(_Display):
     """
     A wrapper class for Finch.Tensor and Finch.SwizzleArray.
 
-    Two constructors are supported: `lvl`+`arr`, or `jl_data`.
+    Three constructors are supported: `lvl` with data, `lvl` with description + `arr`, or `jl_data`.
 
     Parameters
     ----------
     lvl : AbstractLevel, optional
-        Levels description.
+        Levels' description or levels with data. When level's description is passed then
+        `arr` parameter is required. A zero-copy fully dense tensor semantics is available
+        with levels with data constructor (see Examples section).
     arr : ndarray, optional
-        NumPy array that should fill `lvl`.
+        NumPy array that should fill `lvl`. If provided `lvl` should only describe levels.
     jl_data : Finch.SwizzleArray, optional
         Raw Julia object.
     order : str or tuple[int, ...], optional
@@ -44,9 +46,15 @@ class Tensor(_Display):
     --------
     >>> import numpy as np
     >>> import finch
-    >>> x = finch.Tensor(lvl=finch.Dense(finch.Element(0)), arr=np.arange(3))
-    >>> x.todense()
+    >>> t1 = finch.Tensor(lvl=finch.Dense(finch.Element(0)), arr=np.arange(3))
+    >>> t1.todense()
     array([0, 1, 2])
+    >>> # no-copy dense constructor
+    >>> arr2d = np.ones((3, 4))
+    >>> levels = finch.Dense(finch.Dense(finch.Element(1, arr2d.ravel()), 4), 3)
+    >>> t2 = finch.Tensor(lvl=levels)
+    >>> np.shares_memory(t2.todense(), arr2d)
+    True
     """
     row_major = "C"
     column_major = "F"
@@ -56,10 +64,14 @@ class Tensor(_Display):
             raise ValueError("For now only numpy input allowed.")
 
         # constructor for levels description and NumPy array input
-        if lvl is not None and arr is not None and jl_data is None:
-            order = self.preprocess_order(order, arr.ndim)
-            inv_order = tuple(i - 1 for i in jl.invperm(order))
-            self._obj = jl.swizzle(jl.Tensor(lvl._obj, arr.transpose(inv_order)), *order)
+        if lvl is not None and jl_data is None:
+            if arr is not None:
+                order = self.preprocess_order(order, arr.ndim)
+                inv_order = tuple(i - 1 for i in jl.invperm(order))
+                self._obj = jl.swizzle(jl.Tensor(lvl._obj, arr.transpose(inv_order)), *order)
+            else:
+                order = self.preprocess_order(order, self.get_lvl_ndim(lvl._obj))
+                self._obj = jl.swizzle(jl.Tensor(lvl._obj), *order)
         # constructor for a raw julia object
         elif jl_data is not None and lvl is None and arr is None:
             if order is not None:
@@ -107,7 +119,7 @@ class Tensor(_Display):
     def _is_dense(self) -> bool:
         lvl = self._obj.body.lvl
         for _ in self.shape:
-            if not jl.isa(self._obj, jl.Finch.Dense):
+            if not jl.isa(lvl, jl.Finch.Dense):
                 return False
             lvl = lvl.lvl
         return True
@@ -138,6 +150,16 @@ class Tensor(_Display):
             raise ValueError(f"order must be 'C', 'F' or a tuple, but is: {type(order)}.")
 
         return permutation
+
+    @classmethod
+    def get_lvl_ndim(cls, lvl) -> int:
+        ndim = 0
+        for _ in range(np.MAXDIMS):
+            ndim += 1
+            lvl = lvl.lvl
+            if jl.isa(lvl, jl.Finch.Element):
+                break
+        return ndim
 
     def get_order(self, zero_indexing=True) -> tuple[int, ...]:
         order = self._order
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -1,41 +1,39 @@
-from abc import abstractmethod
-from typing import Union
+from typing import Optional, Union
 
 import numpy as np
 import juliacall as jc
 
 from .julia import jl
-
-
-class _Display:
-    def __repr__(self):
-        return jl.sprint(jl.show, self._obj)
-
-    def __str__(self):
-        return jl.sprint(jl.show, jl.MIME("text/plain"), self._obj)
+from .levels import _Display, Dense, Element, Storage
+from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays
 
 
 class Tensor(_Display):
     """
     A wrapper class for Finch.Tensor and Finch.SwizzleArray.
 
-    Three constructors are supported: `lvl` with data, `lvl` with description + `arr`, or `jl_data`.
+    Constructors
+    ------------
+    Tensor(scipy.sparse.spmatrix)
+        Construct a Tensor out of a `scipy.sparse` object. Supported formats are: `COO`, `CSC`, and `CSR`.
+    Tensor(numpy.ndarray)
+        Construct a Tensor out of a NumPy array object. This is a no-copy operation.
+    Tensor(Storage)
+        Initialize a Tensor with a `storage` description. `storage` can already hold data.
+    Tensor(julia_object)
+        Tensor created from a compatible raw Julia object. Must be a `SwizzleArray`.
+        This is a no-copy operation.
 
     Parameters
     ----------
-    lvl : AbstractLevel, optional
-        Levels' description or levels with data. When level's description is passed then
-        `arr` parameter is required. A zero-copy fully dense tensor semantics is available
-        with levels with data constructor (see Examples section).
-    arr : ndarray, optional
-        NumPy array that should fill `lvl`. If provided `lvl` should only describe levels.
-    jl_data : Finch.SwizzleArray, optional
-        Raw Julia object.
-    order : str or tuple[int, ...], optional
-        Order of the tensor. The order numbers the dimensions from the fastest to slowest.
-        The leaf nodes have mode `0` and the root node has mode `n-1`. If the tensor was square
-        of size `N`, then `N .^ order == strides`. Available options are "C" (row-major),
-        "F" (column-major), or a custom order. Default: row-major.
+    obj : np.ndarray or scipy.sparse or Storage or Finch.SwizzleArray
+        Input to construct a Tensor. It's a no-copy operation of for NumPy and SciPy input. For Storage
+        it's levels' description with order. The order numbers the dimensions from the fastest to slowest.
+        The leaf nodes have mode `0` and the root node has mode `n-1`. If the tensor was square of size `N`,
+        then `N .^ order == strides`. Available options are "C" (row-major), "F" (column-major), or a custom
+        order. Default: row-major.
+    fill_value : np.number, optional
+        Only used when `arr : np.ndarray` is passed.
 
     Returns
     -------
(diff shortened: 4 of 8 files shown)


willow-ahrens:
Shouldn't we also have `Tensor(arr)`, where arr is numpy or scipy array and we do a no-copy if possible

willow-ahrens:
like, shouldn't `Tensor(np.Array)` do the no-copy?

hameerabbasi:
> Shouldn't we also have `Tensor(arr)`, where arr is numpy or scipy array and we do a no-copy if possible

Since Python doesn't have function overloading, this is considered a bad practice. I'd recommend `type.from_numpy` and `type.from_scipy_sparse` instead.

mtsokol:
I think no-copy `from_numpy` that gives a dense Tensor also makes sense.

mtsokol:
> like, shouldn't `Tensor(np.Array)` do the no-copy?

Right, that makes more sense (passing `arr` only) than building manually this `lvl` structure.

willow-ahrens:
@hameerabbasi  Could we use a dispatch pattern like:
```
def Tensor(arr):
    arr.to_finch_tensor()
```

Or maybe something like:

```
no_copy_constructors = {numpy.Array: from_numpy, scipy.CSR: from_scipy_CSR}
def Tensor(arr):
    no_copy_constructors[typeof(arr)](arr)
```

willow-ahrens:
@hameerabbasi I mean, we don't do `add_numpy_and_scipy_arrays` or `add_numpy_and_numpy_arrays` separately, I think there's a similar pattern used there right?

hameerabbasi:
There can only be one variant of the `__init__`, so what would need to be done is to "reinterpret" the first argument, with an "incorrect" arg name.

> I mean, we don't do `add_numpy_and_scipy_arrays` or `add_numpy_and_numpy_arrays` separately, I think there's a similar pattern used there right?

Well, scipy arrays just error on mixed ops, unless I'm mistaken.

willow-ahrens:
I think this runs up against https://data-apis.org/array-api/2022.12/design_topics/static_typing.html

willow-ahrens:
though I don't see it mentioned anywhere else in the standard

willow-ahrens:
> There can only be one variant of the __init__, so what would need to be done is to "reinterpret" the first argument, with an "incorrect" arg name.

I think what we're talking about here is just the case where `arr` is not None but `lvl` is indeed None

> Well, scipy arrays just error on mixed ops, unless I'm mistaken.

I feel like there's a real opportunity here to design a simple, extensible system, perhaps with my dictionary approach above, that can easily put all of these different conversions from Numpy, ScipySparse, pydata Sparse and all the formats each supports through the same constructor function, and then just print a helpful error message when we have to densify to convert

hameerabbasi:
> I think what we're talking about here is just the case where `arr` is not None but `lvl` is indeed None

An alternative here would be positional-only arguments, like `def __init__(arr, /, lvl, ...)`. That way the user can't specify `arr` as a keyword and we're free to reinterpret it.

willow-ahrens:
How does the following know which + to call?
```
>>> A = scipy.sparse.random(5, 5, 0.5)
>>> B = numpy.ones((5, 5))
>>> A + B
matrix([[1.        , 1.        , 1.        , 1.        , 1.49019306],
        [1.39644749, 1.        , 1.        , 1.        , 1.87599184],
        [1.85115063, 1.        , 1.97588302, 1.47926403, 1.        ],
        [1.65714027, 1.        , 1.        , 1.69535585, 1.        ],
        [1.        , 1.34376335, 1.00957799, 1.50217521, 1.19109791]])
>>> B + A
matrix([[1.        , 1.        , 1.        , 1.        , 1.49019306],
        [1.39644749, 1.        , 1.        , 1.        , 1.87599184],
        [1.85115063, 1.        , 1.97588302, 1.47926403, 1.        ],
        [1.65714027, 1.        , 1.        , 1.69535585, 1.        ],
        [1.        , 1.34376335, 1.00957799, 1.50217521, 1.19109791]])
```

hameerabbasi:
> How does the following know which + to call?

Right, so there's an `__add__` and `__radd__`, and a sentinel `NotImplemented` is returned whenever we don't know how to handle the specific types.

mtsokol:
image

willow-ahrens:
Hameer, I think I'm misunderstanding your opposition to allowing `Tensor(arr=np.Array)` to use the type of `arr` to figure out which conversion to call. Are you opposed to this? There are many ways we could design it to make it extensible, and maybe there's even an existing convert functionality in python we can use?

willow-ahrens:
Python already has many such methods, like `float(x)` which attempts to convert many different kinds of types to a float.

hameerabbasi:
> Hameer, I think I'm misunderstanding your opposition to allowing `Tensor(arr=np.Array)` to use the type of `arr` to figure out which conversion to call.

I'm not opposing it per se, just saying we should disallow `Tensor(arr=some_array)`, and instead do `Tensor(some_array)` instead. I'm okay with the dictionary approach.

willow-ahrens:
Oh, I see. In that case, what are the full set of constructors you envision?

hameerabbasi:
Here's a small stub:

```python
class Tensor:
    def __init__(self, arr, /, *, levels=None, copy: bool | None=None):
        """Permissive constructor."""
        ...
    
    @classmethod
    def from_numpy(cls, arr, /, copy: bool | None=None):
        """Only accepts NumPy arrays"""
    
    @classmethod
    def from_scipy_sparse(cls, arr, /, copy: bool | None=None):
        """Only accepts SciPy sparse arrays"""
```

`copy=None` would be permissive, zero-copy where possible. `copy=False` prohibits copying, and `copy=True` would force a copy.

willow-ahrens:
I see. One challenge with this approach is that we might want to specify `lvls` and not `arr`

hameerabbasi:
> I see. One challenge with this approach is that we might want to specify `lvls` and not `arr`

Do you see a case where we might need to specify both?

willow-ahrens:
yes. We specify both when we want to copy `arr` into a tensor with storage `lvls`. see https://github.com/willow-ahrens/finch-tensor/blob/183db78289e1d571128586e62ad0dee450d9e865/src/finch/tensor.py#L19-L43

hameerabbasi:
That, I feel is going a bit far -- Like we're overloading one function with way too much functionality, which may be confusing for the average user. However, if we really want to do it in `__init__`, I'd suggest the signature `__init__(self, /, *, arrs=None, levels=None)`, that way we have to be explicit and specify both as keywords.

An alternative would be to make a zero-copy version, then have the format conversion as a separate step: `Tensor(arr).asformat(levels=..., order=...)`.

willow-ahrens:
I want to do whatever feels "pythonic". what example constructors can we draw inspiration from? So far, I have https://numpy.org/doc/stable/reference/generated/numpy.array.html#numpy.array and https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_array.html#scipy.sparse.csr_array, both of which seem to be able to construct from different argument types.

There's also of course https://sparse.pydata.org/en/stable/construct.html, with different constructors for each input, which I'd like to avoid because we can just figure it out from the type of the argument.

Note that in scipy, `csr_array(S)` is `s.to_csr()` when supported.

hameerabbasi:
> I want to do whatever feels "pythonic". what example constructors can we draw inspiration from?

I guess different formats are different types, but I'm guessing here we want to mush them into one type? One option is per-format types/classes, via metaclasses, another is to have one type expose a `.levels` or `.format` property, and do format conversions via [`.asformat(...)`](https://sparse.pydata.org/en/stable/generated/sparse.COO.asformat.html)

As far as Python array API support goes, `.device` can be "overloaded" with functionality to add extra information about the format somehow, see the page on [Array API support](https://data-apis.org/array-api/latest/design_topics/device_support.html). There is precedent for this, e.g. Dask sets the device to a Lazy device if the task graph hasn't been executed.

willow-ahrens:
From the meeting, here's the way we'll implement constructors:
```
import numpy as np

np.empty((11, 22, 33), dtype = np.float64, device = Storage(Dense(Sparse(Element(0.0))), order = “F”))

A = scipy.sparse.csr_matrix(...)

B = Tensor(Storage(Dense(Sparse(Element(0.0)))), order = “F”)
B = Tensor(DenseStorage(order = “F”))
B = Tensor(CSFStorage(order = “F”))
C = Tensor(A) #
reduce(B * C, axis = 1)
@compile function foo(B, C)
    reduce(B * C, axis = 1)
end
D = finch.asarray(A, device = Storage(Dense(Sparse(Element(0.0))), order = “F”))
E = D.to_device(Storage(Dense(Sparse(Element(0.0))), order = “C”)) # Transposes, no zero-copy conversion
F = D.to_device(Storage(Dense(Sparse(Element(0.0))), order = “F”)) # Zero-copy conversion 
F = D.to_device(Storage(Dense(Sparse(Element(0.0))), order = “F”), copy = true) #copies, but same format

sparse.asarray(Storage(Dense(Sparse(Element(0.0))), order = “F”))
Tensor(scipy.sparse.csr_matrix(...))
```

We also agreed to use an ifelse to dispatch on numpy, scipy, etc..., with a `__finch_tensor__` fallback protocol.

mtsokol:

I pushed latest changes from today and this PR currently contains:
1. `finch.Tensor(scipy.sparse.spmatrix)` constructor and `finch.Tensor(numpy.ndarray)` no-copy constructor.
2. Support for no-copy constructor for fully dense Tensors.
3. `levels.py` and `utils.py` files to clean up the code a bit. 
4. `Storage` class for device/formats support.
5. Implementation of `to_device`.

Tomorrow I will bump `Finch.jl` version to make SciPy constructor also non-copying, add `asarray` and some tests to cover all cases from the last comment.

You can take a look at tests to see how it looks right now, e.g.:
```python
a = Tensor(np_array)  # no copy
Tensor(np_array).to_device(Storage(Dense(SparseList(Element(0.0)))))
```

I made `finch.Tensor` a function to allow constructing both Tensors and COO/CSC/CSR classes from the same "constructor". Please share your feedback!

pyproject.toml
```diff
@@ -11,6 +11,7 @@ python = "^3.9"
 juliapkg = "^0.1.10"
 juliacall = "^0.9.15"
 numpy = "^1.19"
+scipy = "^1.7"
```
hameerabbasi: I'd make SciPy optional since it's a bit heavy duty as a dependency.

src/finch/__init__.py
```diff
@@ -24,10 +27,25 @@
...
     "fsprand",
     "permute_dims",
     "COO",
     "CSC",
     "CSF",
     "CSR",
 ]
+
+
+def Tensor(*args, **kwargs):
```
hameerabbasi: This might mess up `isinstance` checks -- why not subclass and override `__init__`?

src/finch/levels.py
```diff
@@ -0,0 +1,80 @@
...
+    def __str__(self):
+        return jl.sprint(jl.show, jl.MIME("text/plain"), self._obj)
+
+
+class AbstractLevel(_Display):
+    pass
+
+
+class Storage:
+    def __init__(self, levels_descr: AbstractLevel, order: OrderType = None) -> None:
```
hameerabbasi: Thanks for the type hints!

src/finch/utils.py
```diff
@@ -0,0 +1,4 @@
+from typing import Union
```
hameerabbasi: Might I suggest `typing.py`?

tests/test_sparse.py
```diff
@@ -23,49 +23,52 @@ def rng():
...
-    coords = np.asarray(
-        [[0, 1, 2, 3, 4], [0, 1, 2, 3, 4]],
-        dtype=np.intp,
+    coords = (
+        np.asarray([0, 1, 2, 3, 4], dtype=np.intp),
+        np.asarray([0, 1, 2, 3, 4], dtype=np.intp),
     )
     data = rng.random(5)
-    scalar = finch.Tensor(finch.Element(2), np.array(2))
+    #scalar = finch.Tensor(finch.Element(2), np.array(2))
```
hameerabbasi: Commented code is usually a bad idea -- I'd remove or restore this before merging.

hameerabbasi (changes requested):
Here are some quick suggestions -- The big ones are about the subclass and the SciPy dependency. I'll leave a more detailed review tomorrow.

pyproject.toml
willow-ahrens: How would we specify the finch constructor for scipy types without depending on scipy

src/finch/__init__.py
willow-ahrens: could someone explain what is going on with TensorWrapper? What is this code meant to do?

willow-ahrens: Why can't we move this code inside the `Tensor.__init__()` method?

willow-ahrens (changes requested):
This looks like a good approach, I think I'm a little confused about some things here though. In particular, I don't quite understand TensorWrapper. Also, I think that `CSR`, `CSC`, etc. should be functions that call `Storage` or `Tensor`, rather than subclasses of `Tensor`, so that the whole call would be `Tensor(CSC())` or `Tensor(COO(4))`, or similar. I think making them subclasses of Tensor is somewhat confusing, because there's no invariant enforcing the levels that they wrap?

pyproject.toml
hameerabbasi: One could specify it as optional so that it's only needed on the relevant code paths (when constructing from SciPy sparse arrays), otherwise not.

willow-ahrens:
I guess, one question I'm left with is why the format aliases aren't just functions. e.g. `finch.csr_storage()` or `finch.csr_tensor()`. We could keep the function signature the same as `Tensor`, but just drop the `lvl` or `order` arguments accordingly.

willow-ahrens:
One question also arises: Is `Storage` serving the same role as `Tensor`? There are some cases where we pass `order` to `Storage` and others where we pass it to `Tensor`. I would expect us to choose between one of the following:

1. 
```
np.empty((11, 22, 33), dtype = np.float64, device = Storage(Dense(Sparse(Element(0.0))), order = “F”))

B = Tensor(Storage(Dense(Sparse(Element(0.0))), order = “F”))
B = Tensor(dense_storage(2, order = “F”))
C = Tensor(A)
D = finch.asarray(A, device = Storage(Dense(Sparse(Element(0.0))), order = “F”))
```

2. 
```
np.empty((11, 22, 33), dtype = np.float64, device = Tensor(Dense(Sparse(Element(0.0))), order = “F”))

B = Tensor(Dense(Sparse(Element(0.0)), order = “F”)
B = dense_tensor(2, order = “F”)
C = Tensor(A)
D = finch.asarray(A, device = Tensor(Dense(Sparse(Element(0.0))), order = “F”))
```

What I'm seeing in the PR feels like a mixture of these two and I'm confused.

src/finch/__init__.py
mtsokol: Here's an explanation: We want to pass SciPy and NumPy objects to `Tensor()` constructor but for `COO`, `CSC`, and `CSR` we have separate classes that inherit from `Tensor`. This is the reason why I added `def Tensor()` instead of putting everything to `Tensor.__init__()` - to be able to return `CSC`, `CSR`, and `COO` instances.

I decided to do it the same way as it's done in `pydata/sparse`: [link](https://github.com/pydata/sparse/blob/6304a0f2df14cb48eabd6f0357396fe870aafe54/sparse/_compressed/compressed.py#L853).

But, sure! If we want to have `CSC`, `COO` and `CSR` just as functions that properly set `Tensor`, then we can keep it all in `Tensor.__init__`. Do you agree?

pyproject.toml
mtsokol: SciPy is only used here for `isinstance` checks so I can make it an optional dependency.

mtsokol:
> There are some cases where we pass order to Storage and others where we pass it to Tensor.

Good point - let's keep `order` as a part of `Storage` only.

pyproject.toml
mtsokol: I removed `scipy` dependency (it's only present in the testing suite)

tests/test_sparse.py
mtsokol: Done!

src/finch/__init__.py
mtsokol: @willow-ahrens I removed `TensorWrapper` and now there's only `Tensor` as before. I also removed `CSC`, `COO` classes, etc. and replaced them with class methods in `Tensor`.

mtsokol:
Ok, no-copy SciPy constructors with tests are completed. I applied most of the review comments, as removing `CSC`, `CSR` classes (the only one left is to make `order` present only in `Storage` object).

mtsokol:
> I guess, one question I'm left with is why the format aliases aren't just functions. e.g. `finch.csr_storage()` or `finch.csr_tensor()`. We could keep the function signature the same as `Tensor`, but just drop the `lvl` or `order` arguments accordingly.

Good point - now it's as `finch.Tensor.construct_csr((data, indices, indptr), shape)` to match SciPy API.

src/finch/tensor.py
```diff
@@ -178,60 +209,105 @@ def permute_dims(self, axes: tuple[int, ...]) -> "Tensor":
...
+        indices = jl.OffByOneVector(indices)
+        indptr = jl.OffByOneVector(indptr)
 
         lvl = jl.Element(dtype(fill_value), data)
         jl_data = jl.swizzle(
             jl.Tensor(
                 jl.Dense(jl.SparseList(lvl, shape[0], indptr, indices), shape[1])
             ),
-            *self.get_permutation(order),
+            *get_permutation_fn(order),
```
willow-ahrens: I think that CSR should always be row-major and CSC should always be column major. I don't think that we should support an order argument when constructing CSR or CSC. that would eliminate the need for this lambda, and also absolve us of having to talk about "Row major CSC"

tests/test_sparse.py
```diff
@@ -23,81 +23,74 @@ def rng():
...
+        np.asarray([0, 1, 2, 3, 4], dtype=np.intp),
     )
     data = rng.random(5)
-    scalar = finch.Tensor(finch.Element(2), np.array(2))
 
-    arr_pydata = sparse.COO(coords, data, shape=(5, 5))
+    arr_pydata = sparse.COO(np.vstack(coords), data, shape=(5, 5))
     arr = arr_pydata.todense()
-    arr_finch = finch.COO(coords, data, shape=(5, 5))
+    arr_finch = finch.Tensor.construct_coo(coords, data, shape=(5, 5))
```
willow-ahrens, line 69: Could we define an alias to make this invocation simpler? i.e. `finch.COO` vs. `finch.Tensor.construct_coo`, so that `construct_coo` is more of an internal syntax?

willow-ahrens (changes requested):
This feels like a much clearer approach, thank you for your efforts! I only ask that we remove the order argument from `construct_csr` and `construct_csc` to simplify things, since those formats have a row-majorness or col-majorness built-in to their definition. I think this would remove the need to use a lambda argument to `_construct_compressed2d_jl_object`.

My comment about renaming the `construct_coo` methods should be considered optional, we could easily do that in a separate PR once things settle.

src/finch/tensor.py
mtsokol: Sure! Done, I removed the lambda argument.

tests/test_sparse.py
mtsokol: Makes sense - let me address it in the follow-up PR!

src/finch/utils.py
mtsokol: Done!

willow-ahrens (approved):

src/finch/tensor.py
```diff
@@ -261,65 +342,7 @@ def permute_dims(x: Tensor, axes: tuple[int, ...]):
...
-        self._obj = jl.SparseCOO[ndim](lvl._obj)
-
-
-class SparseHash(AbstractLevel):
-    def __init__(self, ndim, lvl):
-        self._obj = jl.SparseHash[ndim](lvl._obj)
+def _is_scipy_sparse_obj(x):
+    if hasattr(x, "__module__") and x.__module__.startswith("scipy.sparse"):
+        return True
+    return False
```
```suggestion
    return hasattr(x, "__module__") and x.__module__.startswith("scipy.sparse")
```

src/finch/typing.py
```diff
@@ -0,0 +1,10 @@
+from typing import Union
+
+import juliacall as jc
+import numpy as np
+
+OrderType = Union[str, tuple[int, ...], None]
```
hameerabbasi:
```suggestion
OrderType = Union[Literal["C", "F"], tuple[int, ...], None]
```

src/finch/tensor.py
```diff
@@ -139,13 +164,23 @@ def preprocess_order(
 
         return permutation
 
-    def get_order(self, zero_indexing=True) -> tuple[int, ...]:
+    @classmethod
+    def get_lvl_ndim(cls, lvl: JuliaObj) -> int:
+        ndim = 0
+        for _ in range(np.MAXDIMS):
```
hameerabbasi: 32 dims is a limitation by NumPy, we shouldn't limit ourselves to that necessarily.

```suggestion
        while True:
```

src/finch/tensor.py
```diff
@@ -139,13 +164,23 @@ def preprocess_order(
...
+            ndim += 1
+            lvl = lvl.lvl
```
hameerabbasi, line 158: Before this line, I'd add a `jl.isa` call (and raise on invalidity) to make sure the nesting is done right -- Or is that already handled Julia-side?

hameerabbasi (changes requested):
Minor changes, otherwise LGTM. Thanks for working on this.

src/finch/tensor.py

src/finch/typing.py
mtsokol: Done!

src/finch/tensor.py

hameerabbasi (approved):
Thanks, @mtsokol!

src/finch/tensor.py
mtsokol: I tried in Julia repl and it looks that invalid structure can't be composed.

src/finch/tensor.py
```diff
@@ -1,39 +1,45 @@
...
+        Input to construct a Tensor. It's a no-copy operation of for NumPy and SciPy input. For Tensor
+        `storage` must also be provided.
     jl_data : Finch.SwizzleArray, optional
         Raw Julia object.
-    order : str or tuple[int, ...], optional
-        Order of the tensor. The order numbers the dimensions from the fastest to slowest.
-        The leaf nodes have mode `0` and the root node has mode `n-1`. If the tensor was square
-        of size `N`, then `N .^ order == strides`. Available options are "C" (row-major),
-        "F" (column-major), or a custom order. Default: row-major.
+    storage : Storage, optional
```
willow-ahrens: Small request: Could we make the constructor with Storage take the storage as a positional argument? Could we do the same with jl_data?

willow-ahrens (changes requested):
Would it be possible to move Storage from the keyword to a single positional argument? Consider the difference between

```
   Tensor(Storage(SparseList(Element(0.0)))
```
and

```
   Tensor(arr = None, storage = Storage(SparseList(Element(0.0)))
```

mtsokol:
> Would it be possible to move Storage from the keyword to a single positional argument? Consider the difference between ...

Sure! I can make `Tensor()` accept only one positional argument. But then I need to get rid of `Tensor(arr=Tensor, storage=storage)` constructor, which is used in e.g. `to_device` implementation:

```python
def to_device(self, device: Storage) -> "Tensor":
    return Tensor(arr=self, storage=device)
```

But I can make this constructor a class method, it's mostly used internally here as public API recommends rather `Tensor(arr).to_device(storage)`.

willow-ahrens:
Oh, I see the point of the keyword. I suppose theres also `finch.asarray(arr, device=Storage(...))`. But I agree, the `to_device` approach is a better way to achieve that.

src/finch/tensor.py

mtsokol:
> Oh, I see the point of the keyword. I suppose theres also `finch.asarray(arr, device=Storage(...))`. But I agree, the `to_device` approach is a better way to achieve that.

Sure! Now `Tensor()` accepts only one positional argument `obj` and proceeds according to its type.
For now, instead of `asarray`, I added `_from_other_tensor()` as I want to implement Array API constructor functions separately, if it's Ok.

The PR is again ready from my side!

willow-ahrens (approved):

PR #15, opened 2024-02-28
API: `finch.astype` function

mtsokol:
Hi @willow-ahrens @hameerabbasi!

This PR introduces Array API [`astype` function](https://data-apis.org/array-api/latest/API_specification/generated/array_api.astype.html) (without `device` keyword arg for now).

The test suite will pass once https://github.com/willow-ahrens/Finch.jl/pull/438 is merged and a new release is published.

### DIFF
--- a/src/finch/dtypes.py
+++ b/src/finch/dtypes.py
@@ -1,0 +1,19 @@
+from .julia import jl
+
+
+int_: jl.DataType = jl.Int
+int8: jl.DataType = jl.Int8
+int16: jl.DataType = jl.Int16
+int32: jl.DataType = jl.Int32
+int64: jl.DataType = jl.Int64
+uint: jl.DataType = jl.UInt
+uint8: jl.DataType = jl.UInt8
+uint16: jl.DataType = jl.UInt16
+uint32: jl.DataType = jl.UInt32
+uint64: jl.DataType = jl.UInt64
+float16: jl.DataType = jl.Float16
+float32: jl.DataType = jl.Float32
+float64: jl.DataType = jl.Float64
+complex64: jl.DataType = jl.ComplexF32
+complex128: jl.DataType = jl.ComplexF64
+bool: jl.DataType = jl.Bool
--- a/poetry.lock
+++ b/poetry.lock
@@ -597,4 +597,4 @@ test = ["covdefaults (>=2.3)", "coverage (>=7.2.7)", "coverage-enable-subprocess
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.9"
-content-hash = "71a7cde7f28fc96c03e15c879c3e0b31e825b3496aed9f18a902347c02b4c6a5"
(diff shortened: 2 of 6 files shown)


willow-ahrens:
Awesome! If we wanted to match https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html, we would need casting and order arguments. I do not understand the order argument for np.astype, so I'll need a little help there, but I think the casting argument just errors depending on what kind of type combination is requested, right?

hameerabbasi:
> Awesome! If we wanted to match https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html, we would need casting and order arguments. I do not understand the order argument for np.astype, so I'll need a little help there, but I think the casting argument just errors depending on what kind of type combination is requested, right?

This page is down right now, but it's really what we should be targeting rather than NumPy: https://data-apis.org/array-api/latest/

willow-ahrens:
do we have a snapshot of the page, or is there a markdown somewhere?

hameerabbasi:
It's based off this repo: https://github.com/data-apis/array-api

I've alerted the people responsible to the outage.

willow-ahrens (approved):
LGTM!

src/finch/dtypes.py
```diff
@@ -0,0 +1,27 @@
...
+    (np.float32, jl.Float32),
+    (np.float64, jl.Float64),
+    (np.complex64, jl.ComplexF32),
+    (np.complex128, jl.ComplexF64),
+    (np.bool_, jl.Bool),
+]
+
+dtype_py_to_jl = {pair[0]: pair[1] for pair in _dtype_pairs}
+
+dtype_jl_to_py = {pair[1]: pair[0] for pair in _dtype_pairs}
```
hameerabbasi: We don't really need this -- the Array API standard stipulates that we can use any dtype objects, not necessarily NumPy ones. I think something as simple as

```python
int_ = jl.Int
int8 = jl.Int8
...
```

might suffice.

hameerabbasi (changes requested):
LGTM besides the one comment about extra complexity.

hameerabbasi:
> > Awesome! If we wanted to match https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html, we would need casting and order arguments. I do not understand the order argument for np.astype, so I'll need a little help there, but I think the casting argument just errors depending on what kind of type combination is requested, right?
> 
> This page is down right now, but it's really what we should be targeting rather than NumPy: https://data-apis.org/array-api/latest/

This page is back up, by the way.

hameerabbasi (approved):

src/finch/dtypes.py
mtsokol: That's right - done!

hameerabbasi:
This looks great, thanks, @mtsokol!

PR #17, opened 2024-03-05
API: Lazy API

mtsokol:

This is a draft PR to start the discussion around targeting lazy API in Python. 
`Finch.jl` PR: https://github.com/willow-ahrens/Finch.jl/pull/389

Here's the first idea:
1. ~~`Tensor` class gains a new (private) attribute, `lazy_mode` indicating whether this tensor is taking part in lazy or eager computations.~~
2. ~~Each implemented Array API function (here please look at `def multiply(...)`) wraps input in `LogicTensor` from julia in case it's a raw SwizzleArray.~~
3. Here, as `multiply` is an elem-wise operation we need to ensure Python-style broadcasting (appending missing dimensions to the front). We first invert dims of both arrays, multiply element-wise, and invert back. 
4. If it's an eager mode, we call `Finch.compute` immediately, otherwise the "computation graph" result is returned.
5. ~~The `@compile` (or `@fuse` as we called it originally) sets `lazy_mode` flag to true for all input tensors, calls decorated function and then calls `Finch.compute`. Thanks to this flag all operations within decorated function will be "lazy".~~

I guess `swizzle(in::LogicTensor, dims)` also needs to be implemented.

We should be able to execute the Array API script (https://gist.github.com/mtsokol/f7772df523c05a28116bde75d1d9d3b6) in an eager and `@compile`/lazy mode.

Please share your thoughts!

### DIFF
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -1,0 +1,34 @@
+from functools import wraps
+
+from .julia import jl
+from .tensor import Tensor
+
+
+def compiled(func):
+    @wraps(func)
+    def wrapper_func(*args, **kwargs):
+        new_args = []
+        for arg in args:
+            if isinstance(arg, Tensor) and not jl.isa(arg._obj, jl.Finch.LazyTensor):
+                new_args.append(Tensor(jl.Finch.LazyTensor(arg._obj)))
+            else:
+                new_args.append(arg)
+
+        result = func(*new_args, **kwargs)
+        result_tensor = Tensor(jl.Finch.compute(result._obj))
+
+        return result_tensor
+
+    return wrapper_func
+
+
+def lazy(tensor: Tensor):
+    if tensor.is_computed():
+        return Tensor(jl.Finch.LazyTensor(tensor._obj))
+    return tensor
+
+
+def compute(tensor: Tensor):
+    if not tensor.is_computed():
+        return Tensor(jl.Finch.compute(tensor._obj))
+    return tensor
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, Union
+from typing import Callable, Optional, Union
 
 import numpy as np
 from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
@@ -21,7 +21,7 @@ class Tensor(_Display):
     Tensor(Storage)
         Initialize a Tensor with a `storage` description. `storage` can already hold data.
     Tensor(julia_object)
-        Tensor created from a compatible raw Julia object. Must be a `SwizzleArray`.
+        Tensor created from a compatible raw Julia object. Must be a `SwizzleArray` or `LazyTensor`.
         This is a no-copy operation.
 
     Parameters
@@ -57,8 +57,8 @@ class Tensor(_Display):
     array([[0, 1, 2],
            [3, 4, 5]])
     """
-    row_major = "C"
-    column_major = "F"
+    row_major: str = "C"
+    column_major: str = "F"
 
     def __init__(
         self,
@@ -76,7 +76,9 @@ class Tensor(_Display):
         elif isinstance(obj, Storage):  # from-storage constructor
             order = self.preprocess_order(obj.order, self.get_lvl_ndim(obj.levels_descr._obj))
             self._obj = jl.swizzle(jl.Tensor(obj.levels_descr._obj), *order)
-        elif jl.isa(obj, jl.Finch.SwizzleArray):  # raw-Julia-object constructor
+        elif jl.isa(obj, jl.Finch.Tensor):  # raw-Julia-object constructors
+            self._obj = jl.swizzle(obj, *tuple(range(1, jl.ndims(obj) + 1)))
+        elif jl.isa(obj, jl.Finch.SwizzleArray) or jl.isa(obj, jl.Finch.LazyTensor):
             self._obj = obj
         else:
             raise ValueError(
@@ -84,19 +86,71 @@ class Tensor(_Display):
             )
 
     def __pos__(self):
-        return Tensor(jl.Base.broadcast(jl.seval("+"), self._obj))
+        return self._elemwise_op("+")
+
+    def __neg__(self):
+        return self._elemwise_op("-")
 
     def __add__(self, other):
-        return Tensor(jl.Base.broadcast(jl.seval("+"), self._obj, other._obj))
+        return self._elemwise_op(".+", other)
 
     def __mul__(self, other):
-        return Tensor(jl.Base.broadcast(jl.seval("*"), self._obj, other._obj))
+        return self._elemwise_op(".*", other)
 
     def __sub__(self, other):
-        return Tensor(jl.Base.broadcast(jl.seval("-"), self._obj, other._obj))
+        return self._elemwise_op(".-", other)
 
     def __truediv__(self, other):
-        return Tensor(jl.Base.broadcast(jl.seval("/"), self._obj, other._obj))
+        return self._elemwise_op("./", other)
+
+    def __floordiv__(self, other):
+        return self._elemwise_op(".//", other)
+
+    def __mod__(self, other):
+        return self._elemwise_op("rem", other)
+
+    def __pow__(self, other):
+        return self._elemwise_op(".^", other)
+
+    def __matmul__(self, other):
+        raise NotImplementedError
+
+    def __abs__(self):
+        return self._elemwise_op("abs")
+
+    def __invert__(self):
+        return self._elemwise_op("~")
+
+    def __and__(self, other):
+        return self._elemwise_op("&", other)
+
+    def __or__(self, other):
+        return self._elemwise_op("|", other)
+
+    def __xor__(self, other):
+        return self._elemwise_op("xor", other)
+
+    def __lshift__(self, other):
+        return self._elemwise_op("<<", other)
+
+    def __rshift__(self, other):
+        return self._elemwise_op(">>", other)
+
+    def _elemwise_op(self, op: str, other: Optional["Tensor"] = None) -> "Tensor":
+        if other is None:
+            result = jl.broadcast(jl.seval(op), self._obj)
+        else:
+            axis_x1, axis_x2 = range(self.ndim, 0, -1), range(other.ndim, 0, -1)
+            # inverse swizzle, so `broadcast` appends new dims to the front
+            result = jl.broadcast(
+                jl.seval(op), 
+                jl.permutedims(self._obj, tuple(axis_x1)),
+                jl.permutedims(other._obj, tuple(axis_x2)),
+            )
+            # swizzle back to the original order
+            result = jl.permutedims(result, tuple(range(jl.ndims(result), 0, -1)))
+
+        return Tensor(result)
 
     def __getitem__(self, key):
         if not isinstance(key, tuple):
@@ -106,7 +160,7 @@ class Tensor(_Display):
         key = _add_plus_one(key, self.shape)
 
         result = self._obj[key]
-        if jl.isa(result, jl.Finch.SwizzleArray):
+        if jl.isa(result, jl.Finch.SwizzleArray) or jl.isa(result, jl.Finch.LazyTensor):
             return Tensor(result)
         elif jl.isa(result, jl.Finch.Tensor):
             return Tensor(jl.swizzle(result, *range(1, jl.ndims(result) + 1)))
@@ -141,6 +195,9 @@ class Tensor(_Display):
     @property
     def _order(self) -> tuple[int, ...]:
         return jl.typeof(self._obj).parameters[1]
+
+    def is_computed(self) -> bool:
+        return not jl.isa(self._obj, jl.Finch.LazyTensor)
 
     @classmethod
     def preprocess_order(
@@ -339,7 +396,7 @@ class Tensor(_Display):
         return Tensor(cls.construct_csf_jl_object(arg, shape))
 
 
-def fsprand(*args, order=None):
+def fsprand(*args):
     return Tensor(jl.fsprand(*args))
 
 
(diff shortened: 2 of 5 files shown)


src/finch/tensor.py
```diff
@@ -361,6 +364,49 @@ def astype(x: Tensor, dtype: jl.DataType, /, *, copy: bool = True):
...
+    pass
+
+
+def add(x1: Tensor, x2: Tensor) -> Tensor:
+    pass
+
+
+def multiply(x1: Tensor, x2: Tensor, /) -> Tensor:
+    x1_obj = x1._obj if jl.isa(x1._obj, jl.LogicTensor) else jl.LogicTensor(x1._obj)
+    x2_obj = x2._obj if jl.isa(x2._obj, jl.LogicTensor) else jl.LogicTensor(x2._obj)
```
willow-ahrens: If either is a logic tensor, julia finch should convert the other ones to match.

willow-ahrens: we may want to consider handling "lazy or not lazy" in the julia side, as it's already handled there and the `lazy` boolean doesn't really change how it's implemented in either case. We can keep the boolean, but consider that we don't need to wrap things in lazy or eager here. We also don't need to call compute at the end of this kernel.

willow-ahrens (commented):
This is pretty close to done! To avoid replicating logic, consider letting Finch do some of the wrapping/unwrapping with lazy objects and calling compute.

hameerabbasi:
I'll leave this discussion here on the API design: https://github.com/pydata/sparse/discussions/618

I'd hope/expect that the default mode be lazy, as that'll make things easier for `sparse`.

src/finch/tensor.py
mtsokol: > If either is a logic tensor, julia finch should convert the other ones to match.

Then converting here should happen only when both are just `SwizzleArrays` (an operation performed on two newly created Tensors)

> We also don't need to call compute at the end of this kernel.

The `compute` is called only for the "eager" mode - when `multiply` is called outside of a `@compile` decorated function. Then how do we know when an operation should be eager or lazy? (i.e. inside or outside of `@compile` decorated function)

mtsokol: Or is `LogicTensor` indicating a lazy mode?

willow-ahrens: Logictensor is always lazy, and always propagates, which is why it feels a little redundant to convert things. If both arrays are swizzles and one is lazy, why wasn't it already converted to logictensor? Should I rename logictensor to lazytensor? A few thoughts I had seeing this.

willow-ahrens:
Im not sure what you mean hameer, do you want the default to be lazy all the time or just in a compiled region?

src/finch/tensor.py
mtsokol: If one of the input arrays is lazy then I think whole operation should be lazy - that's fine.

But how about running an operation in eager mode? All inputs must be just swizzles? Then I can remove `lazy_mode` flag from the `Tensor` class here (all inputs are swizzles - eager; at least one `Logictensor` input - lazy).

willow-ahrens: yes, the `Swizzle` and `Tensor` class are eager, but if any input is `LogicTensor` the operation becomes lazy and produces another LogicTensor

willow-ahrens: I'm hoping to get logictensor running today

hameerabbasi:
> Im not sure what you mean hameer, do you want the default to be lazy all the time or just in a compiled region?

Yes, that's desirable. It's easy to go from lazy to eager but not the other way around.

There was a discussion where we noticed `@compile` was rarer and `.compute` was more common among libraries. If we need an eager version, we can always add a one-liner at the top level and fall back to the `@compile` pattern.

willow-ahrens:
I think that we should leave the default eager, that way the user has control of the operation by default, and doesn't need to reason about a lazy api. I think that the two choices of default eager and default lazy are simple to switch between.

willow-ahrens:
if you want to do `.compute`, you  could also always just do 

```
A = LazyTensor(B)
C = A ^2
D = compute(C)
```

willow-ahrens:
that's all `@compile` is sugar for anyway, right?

hameerabbasi:
Okay that sounds sensible. Thanks, I didn't think of a lazy type/marker.

willow-ahrens:
Do we want to skip @compile entirely, since the `LazyTensor` approach is pretty clean? I like it much better tbh.
Compare:
```
A = LazyTensor(B)
C = A ^2
D = compute(C)
```
to
```
@compiled def foo(A):
   C = A^2
   return C
foo(B)
```

hameerabbasi:
Well, to users of existing libraries, `@compile` is much more familiar, and markers are internal sugar.

willow-ahrens:
which libraries use `@compile`, which use markers, and which are lazy by default?

hameerabbasi:
> which libraries use `@compile`, which use markers, and which are lazy by default?

Numba and PyTorch use `@compile`, PyTorch uses markers internally (not user-visible), and JAX is lazy by default.

willow-ahrens:
I see. I guess we can expose both `@compile` and the markers.

hameerabbasi:
Oh, I failed to mention Dask is also lazy, and it uses a `.compute(...)` method. Exposing the markers seems like a good idea, given we may want to tweak the compilation, i.e. choose some scheduling parameters or output formats.

mtsokol:
Thank you both for feedback! In the current version we have two ways for defining lazy computations:
1. `@compile` decorator, that wraps internal `_obj` in `Finch.LogicTensor`
2. New `LogicTensor()` marker to which we can pass `Tensor` (it's just a function which wraps `_obj` in `Finch.LogicTensor`) and `def compute()` for running computations

This way we can have both notations shown in: https://github.com/willow-ahrens/finch-tensor/pull/17#issuecomment-1981375769

- `finch-tensor` isn't aware of any eager or lazy computations - it just passes `_obj` (that can be a SwizzleArray or a LogicTensor) to julia functions. The decision which mode to use, based on inputs, is made on the Finch.jl side.  
- I think we could rename `LogicTensor` to `LazyTensor` as suggested earlier.

willow-ahrens:
Okay, the lazy interface for Finch is set to automerge in https://github.com/willow-ahrens/Finch.jl/pull/389 and is just waiting for tests

willow-ahrens:
I renamed LogicTensor to LazyTensor and made some performance concessions to get things to just work for now.

willow-ahrens:
Mateusz, you'll need to let me know where there are bugs in the system. In particular, the system appears to handle scalars a little strangely right now.

src/finch/compiled.py
```diff
@@ -0,0 +1,35 @@
...
+
+        for arg in args:
+            arg._obj = arg._obj.data
+
+        return result_tensor
+
+    return wrapper_func
+
+
+def LogicTensor(tensor: Tensor):
```
willow-ahrens, line 25: Do we want a LogicTensor in Python, or do we just want the existing Tensor to hold a Julia LogicTensor?

willow-ahrens:
Okay, lazy Finch is released.

mtsokol:
> Okay, lazy Finch is released.

Thank you!

> Mateusz, you'll need to let me know where there are bugs in the system. In particular, the system appears to handle scalars a little strangely right now.

Sure!

willow-ahrens:
Finch 0.6.17 should have the latest fixes to the lazy api, feel free to give it a shot!

mtsokol:
@willow-ahrens Thank you!

willow-ahrens (changes requested):

src/finch/compiled.py
```diff
@@ -0,0 +1,35 @@
...
+
+        result = func(*new_args, **kwargs)
+        result_tensor = Tensor(jl.Finch.compute(result._obj))
+
+        return result_tensor
+
+    return wrapper_func
+
+
+def LazyTensor(tensor: Tensor):
```
mtsokol: @willow-ahrens 
> I don't really understand this function. I thought this function was supposed to wrap the input in LazyTensor but it seems like it's computing the tensor?

src/finch/compiled.py
```diff
@@ -0,0 +1,35 @@
...
+from .julia import jl
+from .tensor import Tensor
+
+
+def compiled(func):
+    @wraps(func)
+    def wrapper_func(*args, **kwargs):
+        new_args = []
+        for arg in args:
+            if isinstance(arg, Tensor) and not jl.isa(arg._obj, jl.Finch.LazyTensor):
```
mtsokol, line 12: @willow-ahrens 
> Let's move this wrapper_func functionality out so that we have a function lazy that wraps tensors in LazyTensor.

src/finch/compiled.py
mtsokol: Hmm, I think `def LazyTensor` is a misleading name here. In `finch-tensor`, `Tensor` class contains `_obj` field which is a raw Julia object. It can be either `Finch.SwizzleArray` or `Finch.LazyTensor`.

We need a function that wraps `_obj` in `Finch.LazyTensor` to make a lazy tensor.

Therefore `def LazyTensor` accepts `Tensor` and wraps `_obj` inside in `Finch.LazyTensor`, that's all.

Maybe I can rename it to `def lazy`, or add a method `Tensor.to_lazy()`, so it makes more sense.

mtsokol: But I agree that currently it's a confusing name.

src/finch/compiled.py
mtsokol: `wrapper_func` needs to stay there, because that's how we define custom decorator in Python.

I renamed `def LazyTensor()` to `def to_lazy()` that wraps `_obj` in `Finch.LazyTensor`.

So for `Tensor` class, the eager mode means that `_obj` field is `Finch.SwizzleArray` and in the lazy mode `_obj` is `Finch.LazyTensor`. Is it Ok?

src/finch/compiled.py
willow-ahrens: I'm in favor of `def lazy`, that would at least match the julia interface

src/finch/compiled.py
willow-ahrens: I think I misunderstood this, yes this is okay!

src/finch/compiled.py
willow-ahrens: `A.to_lazy()` feels like the Array-api style way to do things, and `lazy(A)` feels like the julia-style way to do it. I suppose we also call `A.compute()` or `compute(A)` depending. I sortof prefer `lazy(A)` and `compute(A)` for a frontend interface, but both are good!

src/finch/compiled.py
```diff
@@ -0,0 +1,35 @@
...
+        return result_tensor
+
+    return wrapper_func
+
+
+def to_lazy(tensor: Tensor):
+    if not jl.isa(tensor._obj, jl.Finch.LazyTensor):
+        return Tensor(jl.Finch.LazyTensor(tensor._obj))
+    else:
+        tensor
```
hameerabbasi:
```suggestion
    return tensor
```

Missing `return`, and we can dedent.

src/finch/compiled.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def compute(tensor: Tensor):
+    if not jl.isa(tensor._obj, jl.Finch.LazyTensor):
+        raise ValueError("Input tensor must be a LazyTensor")
```
hameerabbasi:
```suggestion
        return tensor
```

A non-lazy `Tensor` is already computed.

hameerabbasi (changes requested):
A couple of comments -- I agree with @willow-ahrens' thoughts that we should prefer methods over functions, as that's the current state of existing libraries.

src/finch/compiled.py
mtsokol: I think for now I also prefer `lazy(A)` and `compute(A)` versions.

src/finch/compiled.py
mtsokol: Done!

src/finch/compiled.py
mtsokol: Done!

mtsokol:
> we should prefer methods over functions, as that's the current state of existing libraries.

If it's Ok, I would stay with `lazy()` and `compute()` functions for now, as it's closer to Julia - it's just easier to work with for now (function->method is a cosmetic change that I can do anytime later).

src/finch/compiled.py
```diff
@@ -0,0 +1,34 @@
...
+        result = func(*new_args, **kwargs)
+        result_tensor = Tensor(jl.Finch.compute(result._obj))
+
+        return result_tensor
+
+    return wrapper_func
+
+
+def lazy(tensor: Tensor):
+    if not jl.isa(tensor._obj, jl.Finch.LazyTensor):
```
hameerabbasi: I'd suggest adding an `is_lazy()` method to `Tensor` to compute exactly this, and using it here for a cleaner interface. You can also do a function if you prefer.

hameerabbasi (approved):
Just one minor suggestion, otherwise LGTM.

src/finch/compiled.py
mtsokol: @hameerabbasi Did you mean `to_lazy()`? I think `is_lazy` refers to a bool check.

hameerabbasi: I indeed mean a method to check if a given `Tensor` is lazy/computed. We could add an `is_computed()` method instead as well, which I think would be a better idea.

mtsokol: Done!

willow-ahrens: I think this is an instance where the underlying library should also handle this logic. Because e.g. jl.lazy(LazyTensor) is already a no-op and jl.compute(swizzle) should be a no-op if it's not already, can we opt to simplify this logic in a later PR so that we simply always call jl.compute or always call jl.lazy?

willow-ahrens: Because Finch didn't have a lazy interface when we started, I think this project is an opportunity to explore what that interface should look like in both of the libraries, so if it makes sense to add functionality on the python side, it probably makes sense to add it on the julia side too.

mtsokol: That's a good point - `jl.lazy(anything)` should decide whether wrapping happens or not. I can make these adjustment in the next PR.

willow-ahrens: I'm also gonna make a PR to Finch.jl to make sure that if any non-lazy tensors are passed to compute they get returned as-is

Issue #21, opened 2024-03-15
Tracking issue: Add Array API standard support


mtsokol:
Hi @willow-ahrens @hameerabbasi,
This issue is meant to track progress of implementing Array API standard for `finch-tensor`.
I thought that we could try adding short notes to bullet-points, saying which `Finch.jl` functions should be called to implement given entry. I think we already had some ideas during one of our first calls.
Array API: https://data-apis.org/array-api/latest/index.html

A note to those who edit this list in the future: Convert bullets to issues, not subissues, as you complete them. Otherwise they get removed from the list.
### main namespace
- [x] `astype` - https://github.com/willow-ahrens/finch-tensor/pull/15 - eager
- [x] element-wise ops (`add`, `multiply`, `cos`, ...) - https://github.com/willow-ahrens/finch-tensor/pull/17 (partially...)
- [x] reductions (`xp.prod`, `xp.sum`) - `jl.sum` and `jl.prod`, also just `jl.reduce` - https://github.com/willow-ahrens/finch-tensor/pull/17
- [x] `matmul` - implemented with `finch.tensordot` for non-stacked input. Should be rewritten with `jl.mul` / Finch einsum.
- [x] `tensordot` - `finch.tensordot` - https://github.com/willow-ahrens/finch-tensor/pull/22
- [x] `where` - `jl.broadcast(jl.ifelse, cond, a, b)` - https://github.com/willow-ahrens/finch-tensor/pull/30
- [x] #90
- [ ] `take` - `jl.getindex` eager for now
- [x] `nonzero` - this is an eager function, but it is implemented as `ffindnz(arr)` - https://github.com/willow-ahrens/finch-tensor/pull/30
- [x] creation functions: `asarray`, `ones`, `full`, `full_like`, ... - `finch.Tensor` constructor, as well as `jl.copyto!(arr, jl.broadcasted(Scalar(1))`, as well as changing the default of the tensor with `Tensor(Dense(Element(1.0)))`. We may need to distinguish some of these. https://github.com/willow-ahrens/finch-tensor/pull/28, https://github.com/willow-ahrens/finch-tensor/pull/32
- [x] #106
- [ ] set functions: `unique_all`, `unique_counts`, `unique_inverse`, `unique_values` - eager
- [x] `all`, `any`
- [ ] `concat` - eager for now
- [x] #109
- [ ] `flip` -eager for now
- [x] #107
- [ ] `roll` - eager for now
- [x] #108
[...]

PR #22, opened 2024-03-19
API: Implement `tensordot` and `matmul`

mtsokol:
Tracking issue #21 

This WIP PR introduces `finch.tensordot(x1, x2, axes)` and `finch.matmul(x1, x2)`/`x1 @ x2`.
I still need to complete exhaustive testing. 

---

@willow-ahrens I've got one question regarding `matmul`. I just noticed that there's a slight difference between [`tensordot`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.tensordot.html) and [`matmul`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.matmul.html#array_api.matmul) in Array API for >2D input, namely `tensordot` aggregates non-contracted dims from one input and the other, where `matmul` takes two innermost dims for multiplication, and the rest of dims is treated as a stack/batch dimensions.

Here's a NumPy code showing it (described in [notes in docs](https://numpy.org/doc/stable/reference/generated/numpy.matmul.html#numpy-matmul)):
```python
  >>> a = np.ones([9, 5, 7, 4])
  >>> c = np.ones([9, 5, 4, 3])
  >>> np.tensordot(a, c, axes=((-1,), (-2,))).shape
  (9, 5, 7, 9, 5, 3)
  >>> np.matmul(a, c).shape
  (9, 5, 7, 3)
  >>> # n is 7, k is 4, m is 3
```

So I'm not sure I can implement `matmul` as `self.tensordot(other, axes=((-1,), (-2,)))`. WDYT?


willow-ahrens:
Yeah, perhaps we need to define einsum better in Finch and express both `matmul` and `tensordot` in terms of that.

Side note:

I think we should implement `mul!` in Julia for Finch Tensor and call that here.

mtsokol:
@willow-ahrens The PR is ready from my side! (I also want to benchmark these functions)

For now I constrained `matmul` input to 2D tensors, once we have `mul` that can perform matmul on stack of matrices I will come back to this.

src/finch/__init__.py
```diff
@@ -15,13 +15,14 @@
...
     random,
+    tensordot,
+    matmul,
     permute_dims,
     multiply,
     sum,
     prod,
     add,
     subtract,
-    multiply,
```
hameerabbasi, line 24: Is there a reason we're removing `multiply`?

src/finch/tensor.py
```diff
@@ -113,7 +113,11 @@ def __pow__(self, other):
         return self._elemwise_op(".^", other)
 
     def __matmul__(self, other):
-        raise NotImplementedError
+        # TODO: Implement and use mul instead of tensordot
+        # https://github.com/willow-ahrens/finch-tensor/pull/22#issuecomment-2007884763
+        if self.ndim != 2 or other.ndim != 2:
+            raise ValueError(f"Both tensors must be 2-dimensional, but are: {self.ndim} and {other.ndim}.")
```
hameerabbasi: I think this error could be formatted better, it will read like "Both tensors must be 2-dimensional, but are 3 and 4.".
```suggestion
            raise ValueError(f"Both tensors must be 2-dimensional, but: {self.ndim=} and {other.ndim=}.")
```

src/finch/tensor.py
```diff
@@ -463,6 +476,14 @@ def prod(
     return _reduce(x, jl.prod, axis, dtype)
 
 
+def tensordot(x1: Tensor, x2: Tensor, /, *, axes=2) -> Tensor:
+    return x1.tensordot(x2, axes=axes)
```
hameerabbasi: Let's make the function form the main one, and remove the method, since it isn't in the Array API or NumPy API.

hameerabbasi (changes requested):
Couple of API changes, thanks for the excellent work here, @mtsokol.

src/finch/__init__.py
mtsokol: Haha 😄 accidentally `multiply` got on this list twice.

src/finch/tensor.py

src/finch/tensor.py

hameerabbasi (approved):
LGTM, thanks @mtsokol!

PR #23, opened 2024-03-22
API: Add `finch.random` function

mtsokol:
Hi @willow-ahrens @hameerabbasi!

This small PR adds `finch.random` to mimic `sparse.random` (required in benchmarks that I'm running).

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -396,7 +396,15 @@ class Tensor(_Display):
         return Tensor(cls.construct_csf_jl_object(arg, shape))
 
 
-def fsprand(*args):
+def random(shape, density=0.01, random_state=None):
+    args = [*shape, density]
+    if random_state is not None:
+        if isinstance(random_state, np.random.Generator):
+            seed = random_state.integers(np.iinfo(np.int32).max)
+        else:
+            seed = random_state
+        rng = jl.default_rng(seed)
+        args = [rng] + args
     return Tensor(jl.fsprand(*args))
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "finch-tensor"
-version = "0.1.6"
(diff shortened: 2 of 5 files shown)


src/finch/tensor.py
```diff
@@ -396,7 +396,11 @@ def construct_csf(cls, arg: TupleOf3Arrays, shape: tuple[int, ...]) -> "Tensor":
         return Tensor(cls.construct_csf_jl_object(arg, shape))
 
 
-def fsprand(*args):
+def random(shape, density=0.01, random_state=None):
+    args = [*shape, density]
+    if random_state is not None:
+        rng = jl.default_rng(random_state)
+        args = [rng] + args
```
hameerabbasi: Here we should handle the case that `random_state` is a `np.random.Generator` instance, in which case we can just generate some bytes from it to pass to `jl.default_rng`.

mtsokol: Done! `jl.default_rng` only accepts an integer.

willow-ahrens: looks good!

src/finch/tensor.py
```diff
@@ -396,7 +396,15 @@ def construct_csf(cls, arg: TupleOf3Arrays, shape: tuple[int, ...]) -> "Tensor":
         return Tensor(cls.construct_csf_jl_object(arg, shape))
 
 
-def fsprand(*args):
+def random(shape, density=0.01, random_state=None):
+    args = [*shape, density]
+    if random_state is not None:
+        if isinstance(random_state, np.random.Generator):
+            seed = random_state.integers(np.iinfo(np.int32).max)
```
hameerabbasi, line 403: Would a `np.uint64` work here?

hameerabbasi (approved):
LGTM, besides the one comment.

src/finch/tensor.py
mtsokol: Sure it can! I will change that later.

PR #27, opened 2024-03-28
API: Add support and tests for scalars

mtsokol:
Issue: https://github.com/willow-ahrens/finch-tensor/issues/25

This is a new WIP PR that is meant to introduce support and tests for scalars (the previous one was from my fork that I removed).

For scalars we need to decide if we want to use `Scalar` or `SparseScalar` (with customizable default) objects.

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -62,17 +62,22 @@ class Tensor(_Display):
 
     def __init__(
         self,
-        obj: Union[np.ndarray, spmatrix, Storage, JuliaObj],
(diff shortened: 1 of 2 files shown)


hameerabbasi:
I think `Scalar` is better, where `default == value`. There's no point for a scalar where `default != value` because that will only densify what could otherwise be a sparse computation.

willow-ahrens:
Sorry we never got around to this! We're building off of finch-tensor-lite now, so I think this is stale.

PR #28, opened 2024-03-28
API: Add `asarray` function

mtsokol:
Hi @hameerabbasi,

This PR adds [`asarray` Array API function](https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html).

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -4,8 +4,8 @@ import numpy as np
 from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
 
 from .julia import jl
-from .levels import _Display, Dense, Element, Storage
-from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays
+from .levels import _Display, Dense, Element, Storage, DenseStorage, SparseCOO, SparseList
+from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays, DType
 
 
 class Tensor(_Display):
@@ -172,7 +172,7 @@ class Tensor(_Display):
             return result
 
     @property
-    def dtype(self) -> np.dtype:
+    def dtype(self) -> DType:
         return jl.eltype(self._obj.body)
 
     @property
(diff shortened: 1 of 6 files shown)


src/finch/tensor.py
```diff
@@ -412,11 +416,36 @@ def random(shape, density=0.01, random_state=None):
...
 
+def asarray(obj, /, *, dtype=None, format=None):
+    if format not in {"coo", "csr", "csc", "csf", "dense", None}:
+        raise ValueError(f"{format} format not supported.")
+    tensor = obj if isinstance(obj, Tensor) else Tensor(obj)
+
+    if format is not None:
+        order = tensor.get_order()
+        if format == "coo":
+            storage = Storage(SparseCOO(tensor.ndim, Element(tensor.fill_value)), order)
```
hameerabbasi, line 427: Will `Tensor.ndim` work for this one?

src/finch/tensor.py
```diff
@@ -412,11 +416,36 @@ def random(shape, density=0.01, random_state=None):
...
+        elif format == "csr":
+            storage = Storage(Dense(SparseList(Element(tensor.fill_value))), order)
+        elif format == "csc":
+            storage = Storage(Dense(SparseList(Element(tensor.fill_value))), order)
+        elif format == "csf":
+            storage = Storage(Dense(SparseList(SparseList(Element(tensor.fill_value)))), order)
```
hameerabbasi: CSF is NDim-agnostic, it's just `Dense` followed by however many `SparseList` we need.

hameerabbasi (commented):
A couple of questions.

src/finch/tensor.py
mtsokol: Here we need to provide dimensionality of the input tensor. `ndim` is an instance attribute, not class attribute, so `Tensor.ndim` won't work here.

src/finch/tensor.py
mtsokol: Right - updated!

src/finch/tensor.py
hameerabbasi: I mean that would NDim work for a COO-format tensor?

hameerabbasi: Resolved in Slack; `ndim` would do the right thing.

hameerabbasi (approved):

PR #29, opened 2024-04-10
Avoid adding deps every setup

mtsokol:
This PR makes sure that `Finch` dependency is added only once. This should improve startup time (both locally and in benchmarks).


src/finch/julia.py
```diff
@@ -1,6 +1,13 @@
 import juliapkg
 
-juliapkg.add("Finch", "9177782c-1635-4eb9-9bfb-d9dfa25e6bce", version="0.6.20")
+deps = juliapkg.deps.load_cur_deps()
+if not (
+    "packages" in deps and
+    "Finch" in deps["packages"] and
+    deps["packages"]["Finch"]["version"] == "0.6.20"
```
hameerabbasi: I'd simplify this to:
`deps.get(packages, {}).get("Finch", {}).get("version", None) != "0.6.20"`.

mtsokol: Done!

src/finch/julia.py
```diff
@@ -1,6 +1,9 @@
 import juliapkg
 
-juliapkg.add("Finch", "9177782c-1635-4eb9-9bfb-d9dfa25e6bce", version="0.6.20")
+deps = juliapkg.deps.load_cur_deps()
+if deps.get("packages", {}).get("Finch", {}).get("version", None) == "0.6.20":
```
```suggestion
if deps.get("packages", {}).get("Finch", {}).get("version", None) != "0.6.20":
```

src/finch/julia.py
```diff
@@ -1,6 +1,9 @@
 import juliapkg
 
-juliapkg.add("Finch", "9177782c-1635-4eb9-9bfb-d9dfa25e6bce", version="0.6.20")
+deps = juliapkg.deps.load_cur_deps()
+if not deps.get("packages", {}).get("Finch", {}).get("version", None) == "0.6.20":
+    juliapkg.add("Finch", "9177782c-1635-4eb9-9bfb-d9dfa25e6bce", version="0.6.20")
```
hameerabbasi: I'd define `_FINCH_VERSION` and `_FINCH_HASH` and use those -- so it doesn't need updating in two places.

src/finch/julia.py
```diff
@@ -1,6 +1,9 @@
 import juliapkg
 
-juliapkg.add("Finch", "9177782c-1635-4eb9-9bfb-d9dfa25e6bce", version="0.6.20")
+deps = juliapkg.deps.load_cur_deps()
+if not deps.get("packages", {}).get("Finch", {}).get("version", None) == "0.6.20":
```
```suggestion
if deps.get("packages", {}).get("Finch", {}).get("version", None) != "0.6.20":
```

hameerabbasi (changes requested):
Last changes from my side. 😉

hameerabbasi (approved):

willow-ahrens:
Lgtm! Did this end up improving the runtime? The pkg add should already be caching what it can, so I'm curious if this is noticeably faster.

mtsokol:
The startup is faster for sure - `import finch` does run faster if it has already been imported with a given version.

willow-ahrens:
awesome!

PR #30, opened 2024-04-12
API: Implement `where` and `nonzero`

mtsokol:

This PR adds Array API compatible [`where`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.where.html) and [`nonzero`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.nonzero.html) functions.


willow-ahrens (approved):
This one looks good to me!

PR #31, opened 2024-04-15
API: Add `Tensor.to_scipy_sparse`

mtsokol:

This PR adds compat changes to make finch backend compatible with Scipy, sklearn, etc., by adding:
- `SparseArray` marker class that is already used by PyData backend and SciPy, to identify a `pydata/sparse` object.
- `Tensor.from_scipy_sparse()` to convert Finch tensor to SciPy object.
- `Tensor.to_scipy_sparse()` to convert SciPy object back to Finch tensor.

SciPy PR: https://github.com/scipy/scipy/pull/19796

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -5,7 +5,16 @@ from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
 
 from .dtypes import bool as finch_bool
 from .julia import jl
-from .levels import _Display, Dense, Element, Storage, DenseStorage, SparseCOO, SparseList
+from .levels import (
+    _Display,
+    Dense,
+    Element,
+    Storage,
+    DenseStorage,
+    SparseCOO,
+    SparseList,
+    sparse_formats_names,
+)
 from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays, DType
 
 
@@ -314,6 +323,8 @@ class Tensor(_Display, SparseArray):
     def from_scipy_sparse(cls, x) -> "Tensor":
         if not _is_scipy_sparse_obj(x):
             raise ValueError("{x} is not a SciPy sparse object.")
+        if x.format not in ("coo", "csr", "csc"):
+            x = x.asformat("coo")
         return Tensor(x)
 
     @classmethod
@@ -445,8 +456,14 @@ class Tensor(_Display, SparseArray):
             indptr = np.asarray(body.lvl.lvl.ptr) - 1
             sp_class = sp.csr_matrix if order == (1, 0) else sp.csc_matrix
             return sp_class((data, indices, indptr), shape=self.shape)
-
-        raise ValueError("Invalid format. Tensor should be a COO, CSR or CSC.")
+        if (
+            jl.typeof(body.lvl).name.name in sparse_formats_names or
+            jl.typeof(body.lvl.lvl).name.name in sparse_formats_names
+        ):
+            storage = Storage(SparseCOO(self.ndim, Element(self.fill_value)), order)
+            return self.to_device(storage).to_scipy_sparse()
+        else:
+            raise ValueError("Tensor can't be converted to scipy.sparse object.")
 
 
 def random(shape, density=0.01, random_state=None):
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -9,7 +9,13 @@ from .levels import _Display, Dense, Element, Storage, DenseStorage, SparseCOO, 
 from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays, DType
 
 
-class Tensor(_Display):
+class SparseArray:
+    """
+    PyData/Sparse marker class
+    """
+
+
+class Tensor(_Display, SparseArray):
     """
     A wrapper class for Finch.Tensor and Finch.SwizzleArray.
 
@@ -305,6 +311,12 @@ class Tensor(_Display):
         return jl.swizzle(jl.Tensor(lvl._obj), *order)
 
     @classmethod
+    def from_scipy_sparse(cls, x) -> "Tensor":
+        if not _is_scipy_sparse_obj(x):
+            raise ValueError("{x} is not a SciPy sparse object.")
+        return Tensor(x)
+
+    @classmethod
     def _from_scipy_sparse(cls, x) -> JuliaObj:
         if x.format == "coo":
             return cls.construct_coo_jl_object(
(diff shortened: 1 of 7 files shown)


src/finch/tensor.py
```diff
@@ -407,6 +419,35 @@ def construct_csf_jl_object(
...
+            str(jl.typeof(body.lvl).name.name) == "DenseLevel" and
+            str(jl.typeof(body.lvl.lvl).name.name) == "SparseListLevel"
+        ):
+            data = np.asarray(body.lvl.lvl.lvl.val)
+            indices = np.asarray(body.lvl.lvl.idx) - 1
+            indptr = np.asarray(body.lvl.lvl.ptr) - 1
+            sp_class = sp.csr_matrix if order == (1, 0) else sp.csc_matrix
+            return sp_class((data, indices, indptr), shape=self.shape)
+
+        raise ValueError("Invalid format. Tensor should be a COO, CSR or CSC.")
```
willow-ahrens: Let's convert to one of these formats first if we aren't already. If all the levels are dense ish we can convert to dense, otherwise COO is fine. Does Scipy do DCSR?

src/finch/tensor.py
```diff
@@ -304,6 +310,12 @@ def _from_numpy(cls, arr: np.ndarray, fill_value: np.number) -> JuliaObj:
             lvl = Dense(lvl, arr.shape[i])
         return jl.swizzle(jl.Tensor(lvl._obj), *order)
 
+    @classmethod
+    def from_scipy_sparse(cls, x) -> "Tensor":
+        if not _is_scipy_sparse_obj(x):
+            raise ValueError("{x} is not a SciPy sparse object.")
+        return Tensor(x)
```
willow-ahrens, line 328: It looks like Tensor does not attempt to convert scipy formats. Could we similarly copy the format if a zero-copy constructor is not possible here?

willow-ahrens (commented):
This looks good to me. There are implementation details I'm not familiar with (is_scipy_sparse, etc.). My main suggestion is to convert matrices when possible, even if no zero-cost copy approach exists. What do we think?

src/finch/tensor.py
mtsokol: > Let's convert to one of these formats first if we aren't already.

Sure! Let's convert to `COO` then as a default format.

> If all the levels are dense ish we can convert to dense

Dense format won't work, maybe if we have a dense tensor then I can throw an error?

In Scipy sparse, dense input isn't allowed:
```python
import scipy.sparse as sps
sps.linalg.inv(np.ones((3,3)))  # TypeError: Input must be a sparse matrix
```

> Does Scipy do DCSR?

Unfortunately not: https://docs.scipy.org/doc/scipy/reference/sparse.html#sparse-matrix-classes

hameerabbasi: > Dense format won't work, maybe if we have a dense tensor then I can throw an error?

If we get a `np.ndarray` (dense), we can simply use `from_numpy` instead.

src/finch/tensor.py
mtsokol: Currently `finch-tensor` accepts three SciPy formats in a constructor: `COO`, `CSR`, and `CSC`. All three formats are consumed in a zero-copy way.

What do you mean by `similarly copy the format if a zero-copy constructor is not possible`?
To convert SciPy's other formats (DOK, BSR) to COO first with SciPy API?

(I call `Tensor(x)` here because it already knows how to process scipy input. `def from_scipy_sparse` is only for API compliance purposes)

src/finch/tensor.py
willow-ahrens: Oh, I see. If there's no way to convert dense to a scipy format I see that we should error.

src/finch/tensor.py
willow-ahrens: Yes, let's convert SciPy's other formats (DOK, BSR) to COO first with SciPy API.

willow-ahrens: eventually we can convert BSR but we'll need some compiler improvements

src/finch/tensor.py
mtsokol: > If we get a np.ndarray (dense), we can simply use from_numpy instead.

@hameerabbasi I don't think I understand.

`def to_scipy_sparse` is meant to convert a Tensor to SciPy object only. We support `COO`, `CSR`, and `CSC` SciPy formats. In my opinion `to_scipy_sparse` should only return SciPy objects and never return a NumPy array. If Tensor is a dense one, a user shouldn't be able to call `to_scipy_sparse` on it.
I don't think `from_numpy` should be used here.

src/finch/tensor.py
mtsokol: Done!

src/finch/tensor.py

src/finch/tensor.py
```diff
@@ -407,6 +421,41 @@ def construct_csf_jl_object(
...
+            sp_class = sp.csr_matrix if order == (1, 0) else sp.csc_matrix
+            return sp_class((data, indices, indptr), shape=self.shape)
+        if (
+            str(jl.typeof(body.lvl).name.name) == "DenseLevel" and
+            str(jl.typeof(body.lvl.lvl).name.name) == "DenseLevel"
+        ):
+            raise ValueError("Dense Tensor can't be converted to scipy.sparse object.")
+        else:
+            storage = Storage(SparseCOO(self.ndim, Element(self.fill_value)), order)
+            return self.to_device(storage)
```
willow-ahrens: Should this line call to_scipy_sparse recursively?

willow-ahrens: Also, we might want to check first that all of the levels are sparse-like (i.e. we should convert sparseRLE and sparsePoint but not denseRLE)

willow-ahrens (commented):
One small clarification, maybe for my own understanding.

src/finch/tensor.py
mtsokol: Ah, right! I missed a recursive call, now it's:
```python
return self.to_device(storage).to_scipy_sparse()
```

mtsokol: > Also, we might want to check first that all of the levels are sparse-like (i.e. we should convert sparseRLE and sparsePoint but not denseRLE)

So there should be a list of levels that, if present, then the Tensor shouldn't be converted to SparseCOO?

mtsokol: If it's Ok, I can validate it in a follow-up PR.

mtsokol:
The PR is ready from my side!

src/finch/tensor.py
willow-ahrens: The check should be that at least one of the levels (In any position) is from the list:
SparseList
Sparse
SparseHash
SparseCOO
SparseRLE
SparseVBL
SparseBand
SparsePinpoint
SparseInterval

Could you add that check quickly? I think it would be easy to forget

mtsokol: Ah sure, done!

willow-ahrens (approved):
LGTM!

PR #32, opened 2024-04-16
API: Add `eye` function

mtsokol:

This PR adds [`eye`Array API](https://data-apis.org/array-api/latest/API_specification/generated/array_api.eye.html) function. 

@willow-ahrens I think that as a follow-up I could add `format` argument but, for COO:
```julia
using Finch
tns = Tensor(SparseCOO{2}(Element(0)), 3, 4)
@finch begin
    for i=_, j=_
        if i == j
           tns[i, j] += 1.0
        end
    end
end
```
it fails with:
```
ERROR: Finch.FinchProtocolError("cannot modify Finch.VirtualSparseCOOLevel in place (forgot to declare with .= ?)")
```


willow-ahrens:
Declare it first with `tns .= 0`

willow-ahrens:
Do you think we should just add speye and spdiags to finch? I'll work on that this evening.

mtsokol:
`eye` is pretty simple to implement in `finch-tensor`, but sure - Finch functions that follow:
- https://data-apis.org/array-api/latest/extensions/generated/array_api.linalg.diagonal.html
- https://data-apis.org/array-api/latest/API_specification/generated/array_api.eye.html
Would make it even simpler in `finch-tensor`.

mtsokol:
We could have `format=...` argument there? Or always return `coo`

willow-ahrens:
I just realized that speye and spdiags are classic matlab functions, so we should probably also do it in Finch.jl

mtsokol:
> Declare it first with `tns .= 0`

It helped - thank you! For now we have `format=...` keyword argument that supports `dense` and `coo` by building an expression that is evaluated.

willow-ahrens:
Sure! Maybe we should merge this and create an issue in the Finch repo for spdiags and speye

mtsokol:
Sure!

mtsokol:
I created an issue. The PR is ready I think.

willow-ahrens (approved):

Issue #33, opened 2024-04-18
Unexpected `random` behavor


mtsokol:
Hi @willow-ahrens @hameerabbasi,

I just noticed that `finch.random(..., random_state=...)` isn't providing reproducible tensors for a fixed random state.

```python
with sparse.Backend(backend=sparse.BackendType.Finch):
    result = sparse.random((3,3), density=1.0, random_state=0)
    print(result.todense())
```
Should give the same result for multiple runs but it doesn't. I think the reason is that:

```julia
using Random: default_rng
using Finch

rng = default_rng(0)
fsprand(rng, 5, 5)  # different output on multiple runs

rng = default_rng(0)
randn(rng, 4)
```
Also gives different results. The `default_rng` and underlying `TaskLocalRNG` (https://docs.julialang.org/en/v1/stdlib/Random/#Random.TaskLocalRNG) don't accept seed parameter (there could be at least a warning).

In `finch-tensor` in `random` we could replace `default_rng` with `Xoshiro` as it supports `seed` parameter and provides reproducible runs:
```julia
using Random
using Finch

rng = Xoshiro(0)
fsprand(rng, 5, 5)

rng = Xoshiro(0)
fsprand(rng, 5, 5)  # same output
```
WDYT?

willow-ahrens:
I think that https://docs.julialang.org/en/v1/stdlib/Random/#Random.seed! should be able to seed the default_rng, but the stream of numbers it produces will not be the same across julia versions, right? I have no complaints with different RNGs, but I think the default rng is currently xoshiro. The challenge with rng is that often the generator changes with new updates. There's a StableRNG julia package with a stable random number generator across versions. I think we may simply want to avoid testing in a way that requires stability across versions.

willow-ahrens:
You may need to call `seed!(rng, 0)` instead of `rng(0)`

mtsokol:
I don't expect stability across versions for sure. Only for consecutive runs.

> You may need to call `seed!(rng, 0)` instead of `rng(0)`

Sounds good to me - let me update the PR.

mtsokol:
Fixed in https://github.com/willow-ahrens/finch-tensor/pull/34

PR #34, opened 2024-04-18
FIX: Seed RNG correctly

mtsokol:

As described in #33 `default_rng` requires `seed!` for reproducible runs. The test I added fails on `main`.

WDYT?


willow-ahrens (approved):
Looks good to me! Thanks!

PR #35, opened 2024-04-26
API: More elemwise and reduction functions

mtsokol:
Hi @hameerabbasi,

This PR adds another batch of elemwise and reduction functions.


tests/test_ops.py
```diff
@@ -46,9 +46,22 @@ def my_custom_fun(arr1, arr2, arr3):
 
 
 @pytest.mark.parametrize(
-    "meth_name", ["__pos__", "__neg__", "__abs__"],
+    "meth_name", ["log", "log10", "log1p", "log2", "sqrt", "sign", "round", "exp", "expm1", "floor", "ceil"],
```
hameerabbasi: I'd test all the introduced methods

hameerabbasi: `meth_name` is no longer accurate as these are functions.

hameerabbasi (changes requested):

tests/test_ops.py
mtsokol: Done! All new methods are tested and placed in `parametrize`.

hameerabbasi (changes requested):
I'd also add checks for `dtype` and `shape`. Otherwise LGTM.

mtsokol:
> I'd also add checks for `dtype` and `shape`. Otherwise LGTM.

Done! I added checks for these Tensor's attributes.

hameerabbasi (approved):

willow-ahrens:
It occurs to me that perhaps the list of exported names in the module would be a good place to start counting supported functions, is that right?

mtsokol:
We merged `array-api-tests` CI job in `sparse` (https://github.com/pydata/sparse/pull/668) that shows how many functions are already implemented (~46 right now).

PR #36, opened 2024-04-29
Drop python 3.9, Add creation functions.

mtsokol:
Hi @hameerabbasi,

This PR:

- Drops Python 3.9 support, as `sparse` already did it.
- Removes:
  - `poetry.lock` file. I've read that:
    https://stackoverflow.com/a/61076546/6717327
    > The official recommendation of the poetry maintainers is to commit the lockfile if you develop a deployable application (as opposed to a library).
  - and `requirements.txt` (unused)
- Adds a few methods to `Tensor`: `>`, `<`, `==`, etc...
- Adds `bitwise_*` aliases (already tested underlying `Tensor` methods)
- Introduces `full`, `zeros` and `ones` creation functions (We need to fix https://github.com/willow-ahrens/Finch.jl/pull/512 first, as COO is the format used by these functions)


tests/test_sparse.py
```diff
@@ -175,6 +175,25 @@ def test_asarray(arr2d, arr3d, order, format):
         assert_equal(result.todense(), arr)
 
 
+@pytest.mark.parametrize("shape", [10, (3, 3), (2, 1, 5)])
+@pytest.mark.parametrize("dtype", [None, finch.int64, finch.float64])
+def test_full_ones_zeros(shape, dtype):
+    res = finch.full(shape, 2.0, dtype=dtype)
+    assert_equal(res.todense(), np.full(shape, 2.0, dtype))
+    res = finch.full_like(res, 3.0, dtype=dtype)
+    assert_equal(res.todense(), np.full(shape, 3.0, dtype))
```
hameerabbasi: We should test all these functions without `dtype` as well.

hameerabbasi (changes requested):
Thanks for working on this, @mtsokol!

src/finch/tensor.py
```diff
@@ -515,6 +543,39 @@ def asarray(obj, /, *, dtype=None, format=None):
...
+    shape: int | tuple[int, ...],
+    fill_value: jl_dtypes.number,
+    *,
+    dtype: DType | None = None,
+) -> Tensor:
+    if isinstance(shape, int):
+        shape = (shape,)
+    if dtype is not None:
+        fill_value = dtype(fill_value)
+    return Tensor(jl.Tensor(jl.SparseCOO[len(shape)](jl.Element(fill_value), shape)))
```
willow-ahrens: pass in shape as an argument to `jl.Tensor` to initialize it.

willow-ahrens (changes requested):
See my comments regarding segfaults. Also, do we want to add a parameter to specify a format to any of these constructors?

In Julia, there's a distinction between `zeros` (dense) and `spzeros` (sparse CSR). I wonder if we could add a `format` flag, or maybe a `issparse` boolean?

tests/test_sparse.py
mtsokol: Thanks for pointing out! Now one of the `dtype` param values is `None` which is equivalent to not passing it.

src/finch/tensor.py

mtsokol:
> See my comments regarding segfaults. Also, do we want to add a parameter to specify a format to any of these constructors?

Thank you! Now it works as expected.

> In Julia, there's a distinction between `zeros` (dense) and `spzeros` (sparse CSR). I wonder if we could add a `format` flag, or maybe a `issparse` boolean?

Sure! Let me add `format` keyword arg that will accept `dense` and `coo` for now.

willow-ahrens:
Eventually, maybe we can use Hameer's PR for all of these format arguments

src/finch/tensor.py
```diff
@@ -809,6 +883,30 @@ def atan2(x: Tensor, other: Tensor, /) -> Tensor:
...
+def bitwise_right_shift(x1: Tensor, x2: Tensor, /) -> Tensor:
+    return x1 >> x2
+
+
+def bitwise_xor(x1: Tensor, x2: Tensor, /) -> Tensor:
+    return x1 ^ x2
+
+
+def bitwise_invert(x: Tensor, /) -> Tensor:
+    return ~x
```
hameerabbasi: Can we replace these with `from operator import ... [as ...]`?

hameerabbasi: While we're at it, let's do it for all functions where this may apply.

mtsokol: Done! (it reduced a bit of boilerplate code)

hameerabbasi (approved):

src/finch/tensor.py
```diff
@@ -515,6 +543,52 @@ def asarray(obj, /, *, dtype=None, format=None):
...
+    format: str = "coo",
+) -> Tensor:
+    if isinstance(shape, int):
+        shape = (shape,)
+    if dtype is not None:
+        fill_value = dtype(fill_value)
+
+    if format == "coo":
+        return Tensor(jl.Tensor(jl.SparseCOO[len(shape)](jl.Element(fill_value)), *shape))
+    if format == "dense":
```
willow-ahrens, line 560: LGTM! I'll add an issue to update these.

willow-ahrens (approved):
LGTM!

willow-ahrens (commented):
Oh! Also, I just want to say that you can use the Finch.FinchNotation.and and Finch.FinchNotation.or for logical and/or (the operations which return 1 or 0 as opposed to bitwise). Julia doesn't usually name these as functions

PR #39, opened 2024-05-07
API: Add verbose mode and lazy indexing

mtsokol:
Hi @hameerabbasi,

This PR:
- bumps Finch version to the latest release
- adds support for the debug/verbose compute mode
- adds support for lazy indexing, e.g. `tns[:, None, :]`, that adds a new singleton dimension at that position. There's an issue with the rightmost `None` indexing (https://github.com/willow-ahrens/Finch.jl/issues/528) but I think we can merge it here and I will update tests once it's solved in Finch.jl.
-

### DIFF
--- a/src/finch/julia.py
+++ b/src/finch/julia.py
@@ -1,6 +1,6 @@
 import juliapkg
 
-_FINCH_VERSION = "0.6.23"
+_FINCH_VERSION = "0.6.24"
 _FINCH_HASH = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
 
 deps = juliapkg.deps.load_cur_deps()
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -883,7 +883,7 @@ def _add_plus_one(key: tuple, shape: tuple[int, ...]) -> tuple:
             idx = normalize_axis_tuple(idx, size)
             new_key += (jl.Vector([i + 1 for i in idx]),)
         elif idx is None:
-            raise IndexError("'None' in the index is supported only in lazy indexing")
+            raise IndexError("`None` in the index is supported only in lazy indexing")
         else:
             new_key += (idx,)
     return new_key
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -197,9 +197,15 @@ class Tensor(_Display, SparseArray):
     def __getitem__(self, key):
         if not isinstance(key, tuple):
             key = (key,)
-        key = _expand_ellipsis(key, self.shape)
-        key = _add_missing_dims(key, self.shape)
-        key = _add_plus_one(key, self.shape)
+
+        if None in key:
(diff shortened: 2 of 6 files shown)


src/finch/julia.py
```diff
@@ -1,6 +1,6 @@
 import juliapkg
 
-_FINCH_VERSION = "0.6.23"
+_FINCH_VERSION = "0.6.24"
 _FINCH_HASH = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
```
hameerabbasi, line 4: Does this hash need updating?

src/finch/tensor.py
```diff
@@ -877,7 +883,7 @@ def _add_plus_one(key: tuple, shape: tuple[int, ...]) -> tuple:
             idx = normalize_axis_tuple(idx, size)
             new_key += (jl.Vector([i + 1 for i in idx]),)
         elif idx is None:
-            raise IndexError("'None' in the index key isn't supported")
+            raise IndexError("'None' in the index is supported only in lazy indexing")
```
```suggestion
            raise IndexError("`None` in the index is supported only in lazy indexing")
```

tests/test_indexing.py
```diff
@@ -49,7 +58,9 @@ def test_indexing_2d(arr2d, index, order):
         (slice(None, None, -1), slice(None, None, -1), slice(None, None, -1)),
         (slice(None, -1, 1), slice(-1, None, -1), slice(4, 1, -1)),
         (-1, 0, 0), (0, -1, -2), ([1, 2], 0, slice(3, None, -1)),
-        (0, slice(1, 0, -1), 0),
+        (0, slice(1, 0, -1), 0), (slice(None), None, slice(None), slice(None)),
+        # https://github.com/willow-ahrens/Finch.jl/issues/528
+        # (slice(None), slice(None), slice(None), None),
```
hameerabbasi: Let's add this as an `xfail`-ed test instead.

hameerabbasi (changes requested):

src/finch/julia.py
mtsokol: No, hash stays the same, it's associated with the package.

mtsokol:
@hameerabbasi FYI there's one regression that causes one test failure: https://github.com/willow-ahrens/Finch.jl/issues/524

hameerabbasi:
> @hameerabbasi FYI there's one regression that causes one test failure: [willow-ahrens/Finch.jl#524](https://github.com/willow-ahrens/Finch.jl/issues/524)

Let's also add it to `xfail` for now.

tests/test_indexing.py
mtsokol: Done!

src/finch/tensor.py

hameerabbasi (approved):

hameerabbasi:
Thanks for getting to this so quick, @mtsokol! In it goes.

PR #43, opened 2024-05-09
Verify SciPy canonical input

mtsokol:
Hi @hameerabbasi,

When consuming SciPy sparse object, Finch expects it to be in a canonical format. When it's not true then array isn't loaded properly.

Here I verify that input is in canonical format and raise a proper warning.

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -1,5 +1,6 @@
 import builtins
 from typing import Any, Callable, Optional, Iterable, Literal
+import warnings
 
 import numpy as np
 from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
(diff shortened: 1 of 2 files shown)


willow-ahrens:
does "canonical" mean "sorted"? just curious. thanks!

mtsokol:
> does "canonical" mean "sorted"? just curious. thanks!

[Yes!](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.has_canonical_format.html) `finch-tensor` requires it. I realized that creating `SparseCOO` with coords arrays that aren't sorted results in a broken Tensor (e.g. when densifying (copying to a dense Tensor) the resulting tensor is incorrect).

hameerabbasi:
> does "canonical" mean "sorted"? just curious. thanks!

Almost, sorted + no duplicate entries. 😄

src/finch/tensor.py
```diff
@@ -357,12 +358,21 @@ def _from_numpy(cls, arr: np.ndarray, fill_value: np.number) -> JuliaObj:
...
+        if x.format not in ("coo", "csr", "csc"):
+            x = x.asformat("coo")
+        if not x.has_canonical_format:
+            warnings.warn(
+                "SciPy sparse input must be in a canonical format. "
+                "Calling `sum_duplicates`."
+            )
+            x = x.copy()
+            x.sum_duplicates()
+            assert x.has_canonical_format
```
hameerabbasi, line 374: Can we have a `copy : bool | None` kwarg that can ensure zero-copy when it needs to be ensured, and pass that in via `asarray`.

mtsokol: Can I add `copy` argument in a follow-up PR? I started working on it and it requires a bit more changes in `tensor.py`. For now I would like to be sure that SciPy sparse objects are always loaded correctly. Is it Ok?

hameerabbasi: Sure, I've created #46 to track that.

hameerabbasi (approved):

PR #44, opened 2024-05-09
Use `jl.permutedims` in `permute_dims`

mtsokol:
Hi @hameerabbasi @willow-ahrens,

This small fix makes `permute_dims` use `jl.permutedims` instead of `jl.swizzle`.

It's needed to make transpositions work in the lazy mode, because for `swizzle(lazy(tensor), ...)` we get:

```
ERROR: MethodError: swizzle(::Finch.LazyTensor{Float64, 3}, ::Int64, ::Int64, ::Int64) is ambiguous.

Candidates:
  swizzle(body, dims::Int64...)
    @ Finch ~/JuliaProjects/Finch.jl/src/tensors/combinators/swizzle.jl:80
  swizzle(arr::Finch.LazyTensor, dims...)
    @ Finch ~/JuliaProjects/Finch.jl/src/interface/lazy.jl:63
```

and `permutedims` doesn't raise it.

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -323,7 +323,7 @@ class Tensor(_Display, SparseArray):
 
     def permute_dims(self, axes: tuple[int, ...]) -> "Tensor":
         axes = tuple(i + 1 for i in axes)
-        new_obj = jl.swizzle(self._obj, *axes)
+        new_obj = jl.permutedims(self._obj, axes)
         new_tensor = Tensor(new_obj)
         return new_tensor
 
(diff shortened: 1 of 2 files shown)


willow-ahrens:
I think we should fix that ambiguity error too though. I'll add the issue to finch.jl

mtsokol:
Sure! For now it's a small fix here in `finch-tensor`.

hameerabbasi (changes requested):
Can we add a regression test? Something which ensures `permute_dims` will always compile?

mtsokol:
> Can we add a regression test? Something which ensures `permute_dims` will always compile?

Sure! Done! I extended `test_permute_dims` so it also runs lazy mode (it passes here and fails in a main branch).

hameerabbasi (approved):

hameerabbasi:
Thanks for the diagnosis and the work put towards fixing this, @mtsokol!

PR #45, opened 2024-05-10
`copy` keyword for `asarray`, `Tensor`, and `astype` 

### LINKED ISSUE #46: Add `copy: bool | None` argument for `asarray`
_Originally posted by @mtsokol in https://github.com/willow-ahrens/finch-tensor/pull/43#discussion_r1596473010_

mtsokol:
WIP copy keyword for `asarray`, `Tensor`, and `astype`.

Fixes #46

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -1,5 +1,6 @@
 import builtins
 from typing import Any, Callable, Optional, Iterable, Literal
+import warnings
 
 import numpy as np
 from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
@@ -91,7 +92,9 @@ class Tensor(_Display, SparseArray):
         copy: bool | None = None,
     ):
         if isinstance(obj, (int, float, complex, bool, list)):
-            obj = np.array(obj, copy=copy)
+            if copy is False:
+                raise ValueError("copy=False isn't supported for scalar inputs and Python lists")
+            obj = np.asarray(obj)
         if fill_value is None:
             fill_value = 0.0
 
@@ -424,12 +427,12 @@ class Tensor(_Display, SparseArray):
         fill_value: np.number | None = None,
         copy: bool | None = None,
     ) -> JuliaObj:
-        if copy is False and not (x.has_canonical_format and x.format in ("coo", "csr", "csc")):
+        if copy is False and not (x.format in ("coo", "csr", "csc") and x.has_canonical_format):
             raise ValueError("Unable to avoid copy while creating an array as requested.")
-        if copy or not x.has_canonical_format:
-            x = x.copy()
         if x.format not in ("coo", "csr", "csc"):
             x = x.asformat("coo")
+        if copy:
+            x = x.copy()
         if not x.has_canonical_format:
             x.sum_duplicates()
             assert x.has_canonical_format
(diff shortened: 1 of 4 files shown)


src/finch/tensor.py
```diff
@@ -84,24 +87,30 @@ def __init__(
         /,
         *,
         fill_value: np.number = 0.0,
+        copy: bool | None = None,
     ):
         if isinstance(obj, (int, float, complex, bool)):
-            obj = np.array(obj)
-
+            obj = np.asarray(obj)
```
hameerabbasi, line 97:
```suggestion
            obj = np.asarray(obj, copy=copy)
```

hameerabbasi (commented):
Does Julia not have a way to do a deep copy?

src/finch/tensor.py
mtsokol: `copy` keyword in `asarray` isn't released yet. I will be shipped in NumPy 2.0.

src/finch/tensor.py
```diff
@@ -1,5 +1,6 @@
 import builtins
 from typing import Any, Callable, Optional, Iterable, Literal
+import warnings
```
hameerabbasi, line 3: We don't use `warnings` anymore do we?

mtsokol: There's a warning in `def reshape`, so we need to keep that import.

mtsokol:
> Does Julia not have a way to do a deep copy?

Hmm let's discuss it during Finch meeting. But I would leave copying logic in Python and not rely on Julia copying, considering that we already have no-copy logic for NumPy and SciPy inputs in `finch-tensor`. Let's not disperse it over Python and Julia side.

hameerabbasi:
> > Does Julia not have a way to do a deep copy?
> 
> Hmm let's discuss it during Finch meeting. But I would leave copying logic in Python and not rely on Julia copying, considering that we already have no-copy logic for NumPy and SciPy inputs in `finch-tensor`. Let's not disperse it over Python and Julia side.

Would you like to implement it later, or in this PR? If the former, let's raise an issue to keep track.

src/finch/tensor.py
```diff
@@ -615,13 +643,23 @@ def random(shape, density=0.01, random_state=None):
...
+) -> Tensor:
     if format not in {"coo", "csr", "csc", "csf", "dense", None}:
         raise ValueError(f"{format} format not supported.")
     _validate_device(device)
-    tensor = obj if isinstance(obj, Tensor) else Tensor(obj, fill_value=fill_value)
-
+    tensor = obj if isinstance(obj, Tensor) else Tensor(obj, fill_value=fill_value, copy=copy)
     if format is not None:
+        if copy is False:
+            raise ValueError("Unable to avoid copy while creating an array as requested.")
```
hameerabbasi, line 662: If the format is the same as the current format, we shouldn't hit this right?

mtsokol:
> Would you like to implement it later, or in this PR? If the former, let's raise an issue to keep track.

Let's implement it in a follow-up PR. You mean to allow copying Julia input, right? Like `finch.Tensor(julia_obj, copy=True)`

src/finch/tensor.py
mtsokol: But we don't really have a string representation of a format or a format property in the `Tensor` class. So there's no way checking if the current format is e.g. `csf` and I would prefer to avoid parsing Julia's `_obj` repr if it matches some regex - it appears counterproductive to me.

hameerabbasi (approved):
I created #66 and #67 for the remaining issues, this looks good to go for me.

PR #48, opened 2024-05-13
Array API support - continuation

mtsokol:
Hi @hameerabbasi,

This PR add a couple of simple Array API functions:
```
    "finfo",  # ported from numpy
    "iinfo",  # ported from numpy
    "isnan",  # the rest is just element-wise functions
    "isinf",
    "isfinite",
    "equal",
    "not_equal",
    "less",
    "less_equal",
    "greater",
    "greater_equal",
    "square",
    "logaddexp",
    "logical_and",
    "logical_or",
    "logical_xor",
    "trunc",
```
Adds `__bool__`, etc. to `Tensor` class.

It also introduces a few modifications to correctly run `array-api-test` suite.

---

The tricky part is `reshape` which we don't have right now but it's required by most of `array-api-tests` suites to even run them.
Maybe one option could be another env var `TEST_MODE` that when set enables temporary implementation (that moves to numpy, reshapes, and moves back to Tensor). Otherwise `NotImplementedError` is raised. WDYT?

### DIFF
--- a/src/finch/__init__.py
+++ b/src/finch/__init__.py
@@ -15,8 +15,21 @@ from operator import (
     and_ as bitwise_and,
     lshift as bitwise_left_shift,
     rshift as bitwise_right_shift,
+    eq as equal,
+    ne as not_equal,
+    lt as less,
+    le as less_equal,
+    gt as greater,
+    ge as greater_equal,
+    mod as remainder,
 )
-
+from numpy import (
+    e as e,
+    pi as pi,
+    inf as inf,
+    nan as nan,
+    newaxis as newaxis,
+)
 from .levels import (
     Dense,
     Element,
@@ -77,6 +90,16 @@ from .tensor import (
     ones_like,
     zeros,
     zeros_like,
+    isnan,
+    isfinite,
+    isinf,
+    reshape,
+    square,
+    logaddexp,
+    trunc,
+    logical_and,
+    logical_or,
+    logical_xor,
 )
 from .compiled import (
     lazy,
@@ -100,6 +123,9 @@ from .dtypes import (
     complex64,
     complex128,
     bool,
+    finfo,
+    iinfo,
+    can_cast,
 )
 
 __all__ = [
@@ -195,4 +221,31 @@ __all__ = [
     "bitwise_right_shift",
     "bitwise_xor",
     "bitwise_invert",
+    "finfo",
+    "iinfo",
+    "isnan",
+    "isinf",
+    "isfinite",
+    "reshape",
+    "equal",
+    "not_equal",
+    "less",
+    "less_equal",
+    "greater",
+    "greater_equal",
+    "square",
+    "logaddexp",
+    "logical_and",
+    "logical_or",
+    "logical_xor",
+    "trunc",
+    "e",
+    "pi",
+    "inf",
+    "nan",
+    "newaxis",
+    "can_cast",
+    "remainder",
 ]
+
+__array_api_version__: str = "2023.12"
--- a/src/finch/dtypes.py
+++ b/src/finch/dtypes.py
@@ -1,3 +1,5 @@
+import builtins
+
 import numpy as np
 
 from .julia import jl
@@ -31,6 +33,7 @@ jl_to_np_dtype = {
     int16: np.int16,
     int32: np.int32,
     int64: np.int64,
+    uint: np.uint,
     uint8: np.uint8,
     uint16: np.uint16,
     uint32: np.uint32,
@@ -40,6 +43,17 @@ jl_to_np_dtype = {
     float64: np.float64,
     complex64: np.complex64,
     complex128: np.complex128,
-    bool: np.bool_,
+    bool: builtins.bool,
     None: None,
 }
+
+def finfo(dtype):
+    return np.finfo(jl_to_np_dtype[dtype])
+
+def iinfo(dtype):
+    return np.iinfo(jl_to_np_dtype[dtype])
+
+def can_cast(from_, to, /) -> builtins.bool:
+    if hasattr(from_, "dtype"):
+        from_ = from_.dtype
+    return np.can_cast(jl_to_np_dtype[from_], jl_to_np_dtype[to])
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -6,6 +6,7 @@ import numpy as np
 from numpy.core.numeric import normalize_axis_index, normalize_axis_tuple
 
 from . import dtypes as jl_dtypes
+from .errors import PerformanceWarning
 from .julia import jl
 from .levels import (
     _Display,
@@ -86,7 +87,7 @@ class Tensor(_Display, SparseArray):
         *,
         fill_value: np.number = 0.0,
     ):
-        if isinstance(obj, (int, float, complex, bool)):
+        if isinstance(obj, (int, float, complex, bool, list)):
             obj = np.array(obj)
 
         if _is_scipy_sparse_obj(obj):  # scipy constructor
@@ -104,9 +105,12 @@ class Tensor(_Display, SparseArray):
             self._obj = jl.swizzle(obj, *tuple(range(1, jl.ndims(obj) + 1)))
         elif jl.isa(obj, jl.Finch.SwizzleArray) or jl.isa(obj, jl.Finch.LazyTensor):
             self._obj = obj
+        elif isinstance(obj, Tensor):
+            self._obj = obj._obj
         else:
             raise ValueError(
-                "Either `arr`, `storage` or a raw julia object should be provided."
+                "Either scalar, numpy, scipy.sparse or a raw julia object should "
+                f"be provided. Found: {type(obj)}"
             )
 
     def __pos__(self):
@@ -188,17 +192,36 @@ class Tensor(_Display, SparseArray):
         if other is None:
             result = jl.broadcast(jl.seval(op), self._obj)
         else:
-            axis_x1, axis_x2 = range(self.ndim, 0, -1), range(other.ndim, 0, -1)
+            if not np.isscalar(other):
+                other = jl.permutedims(other._obj, tuple(range(other.ndim, 0, -1)))
             # inverse swizzle, so `broadcast` appends new dims to the front
             result = jl.broadcast(
-                jl.seval(op),
-                jl.permutedims(self._obj, tuple(axis_x1)),
-                jl.permutedims(other._obj, tuple(axis_x2)),
+                jl.seval(op), jl.permutedims(self._obj, tuple(range(self.ndim, 0, -1))), other
             )
             # swizzle back to the original order
             result = jl.permutedims(result, tuple(range(jl.ndims(result), 0, -1)))
 
         return Tensor(result)
+
+    def __bool__(self):
+        return self._to_scalar(bool)
+
+    def __float__(self):
+        return self._to_scalar(float)
+
+    def __int__(self):
+        return self._to_scalar(int)
+
+    def __index__(self):
+        return self._to_scalar(int)
+
+    def __complex__(self):
+        return self._to_scalar(complex)
+
+    def _to_scalar(self, builtin):
+        if self.ndim != 0:
+            raise ValueError(f"{builtin} can be computed for one-element tensors only.")
+        return builtin(self.todense().flatten()[0])
 
     def __getitem__(self, key):
         if not isinstance(key, tuple):
@@ -349,7 +372,8 @@ class Tensor(_Display, SparseArray):
             dtype == np.bool_
         ):  # Fails with: Finch currently only supports isbits defaults
             dtype = jl_dtypes.bool
-        lvl = Element(dtype(fill_value), arr.reshape(-1, order=order_char))
+        fill_value = dtype(fill_value)
+        lvl = Element(fill_value, arr.reshape(-1, order=order_char))
         for i in inv_order:
             lvl = Dense(lvl, arr.shape[i])
         return jl.swizzle(jl.Tensor(lvl._obj), *order)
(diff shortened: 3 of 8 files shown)


src/finch/tensor.py
```diff
@@ -188,18 +191,37 @@ def _elemwise_op(self, op: str, other: Optional["Tensor"] = None) -> "Tensor":
...
+
+    def __index__(self):
+        return self._to_scalar(int)
+
+    def __complex__(self):
+        return self._to_scalar(complex)
+
+    def _to_scalar(self, builtin):
+        if self.size != 1 or self.shape != ():
+            raise ValueError(f"{builtin} can be computed for one-element tensors only.")
```
```suggestion
        if self.ndim != 0:
            raise ValueError(f"{builtin} can be computed for scalar tensors only.")
```

src/finch/tensor.py
```diff
@@ -587,25 +609,37 @@ def asarray(obj, /, *, dtype=None, format=None):
         return tensor
 
 
+def reshape(x: Tensor, /, shape: tuple[int, ...], *, copy: bool | None = None) -> Tensor:
+    # Only to run array-api-tests that require it for multiple tests. Maybe raise a warning
+    # that it densifies? Or enable when some env var is present?
```
hameerabbasi: Let's not densify, `scipy.sparse` made that mistake. It was a source for performance cliffs.

src/finch/tensor.py
```diff
@@ -587,25 +609,37 @@ def asarray(obj, /, *, dtype=None, format=None):
...
+    if format == "coo" and shape != ():
         return Tensor(
-            jl.Tensor(jl.SparseCOO[len(shape)](jl.Element(fill_value)), *shape)
+            jl.Tensor(jl.SparseCOO[len(shape)](jl.Element(dtype(fill_value))), *shape)
         )
-    if format == "dense":
-        return Tensor(np.full(shape, fill_value, dtype=jl_dtypes.jl_to_np_dtype[dtype]))
-    raise ValueError(f"{format} format not supported.")
+    else:
+        return Tensor(np.full(shape, fill_value, dtype=dtype))
```
hameerabbasi, line 646: This returns a fully dense tensor. Can we return an empty tensor with the right fill-value?

hameerabbasi (changes requested):
We also need tests for everything here.

src/finch/__init__.py
```diff
@@ -100,8 +112,15 @@
...
     complex128,
     bool,
+    finfo,
+    iinfo,
 )
 
+logical_and = bitwise_and
+logical_or = bitwise_or
+logical_xor = bitwise_xor
+logical_not = bitwise_invert
```
hameerabbasi: These are distinct -- `logical_*` casts to `bool` before any ops. `bitwise_*` performs bit-wise ops with integers.

willow-ahrens: In Finch, logical and is the `Finch.and` function, while bitwise and is `&`

src/finch/tensor.py
willow-ahrens: only if the fill value is "special" right? we wouldn't want to compile a ton of kernels on accident.

mtsokol: There are two formats supported by `full` here: `coo` and `dense`. The `else` branch is for `dense` case so I think we should return an actual dense tensor. Or do we want a COO format?

willow-ahrens: in Julia and Matlab we have `spzeros` to explicitly ask for a sparse array of zeros, as opposed to `zeros` which returns a dense array. I think this would be resolved if we simply require the user to type `dense` or `coo` and the else case is an error. That way, the user can't make a dense tensor on accident.

willow-ahrens: Oh, I see, we check it at the top. In that case, I think this is fine.

hameerabbasi: We could run into a performance cliff if we choose the incorrect default either way. I think the right solution would be the middle solution with a dynamic fill-value, treating it as a runtime constant rather than a compile time constant, except for absorbing or identity elements.

willow-ahrens: Finch doesn't support dynamic sparse fill values yet.

willow-ahrens: I could add it with the ISO level?

willow-ahrens: but it's not a common thing to do

hameerabbasi: Hmm. 🤔 Not an easy decision either way. Let's pick any solution here and create a Todo comment and an issue to fix it up later?

willow-ahrens: I'm okay with the code as written, though I think it could be improved to make it clearer that dense is not the default but an intentional user decision.

hameerabbasi: The issue with that is we're introducing finch-specific extensions to the API, and they're highly unlikely to get used.

willow-ahrens: which extension here is finch-specific? `full`? or the `format` keyword?

mtsokol: I think nothing is finch-specific. `full` follows Array API with and additional keyword only `format` argument that is also present in `full` in Numba backend.

willow-ahrens: If you could change the `else` to an `elif format == dense`, I think I approve this PR.

src/finch/tensor.py
willow-ahrens: Instead, can we check that all dimensions, if present, are of size 1?

i.e. `if all(dim == 1 for dim in self.shape):`

willow-ahrens: Similarly, could we compute the scalar value by indexing all dimensions with 1?

src/finch/tensor.py
willow-ahrens: I'm on it, perhaps we can leave this for testing for now?

willow-ahrens (changes requested):
Looks good! Just a few tweaks to bitwise ops, the scalar changes, and the one change in `full`

src/finch/tensor.py
hameerabbasi: `format` is sparse-specific.

src/finch/tensor.py
mtsokol: Sure! Thank you!

src/finch/tensor.py
hameerabbasi: That would be in contrast to the Array API and NumPy.

mtsokol: Right, I think we can stick to Array API standard, and support only 0-D arrays, like the suggested change states:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__bool__.html
https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__int__.html

NumPy already warns that >0D is deprecated:
```
In [43]: np.array([0]).__int__()
<ipython-input-43-6afc8c458142>:1: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
  np.array([0]).__int__()
```

willow-ahrens: sounds good!

src/finch/__init__.py
mtsokol: I found:
```
Finch.and
Finch.or
Finch.xor
```
but I think `invert` or `not` is missing, right? (because it's tricky as it changes the fill value?)

src/finch/tensor.py
mtsokol: I updated `if` statement and kept the `format` keyword for this PR but if you think it should be dropped as it's sparse-specific, let me know - I will keep only COO then. (but Numba backend also has `format` keyword for these functions and accepts for example "gcxs")

src/finch/tensor.py

mtsokol:
The PR is ready from my side - I added tests and addressed all comments.

src/finch/tensor.py
willow-ahrens: No, I think adding a format keyword is becoming a good, uniform way to incorporate sparsity into these functions without changing the signatures too much, let's keep doing it!

src/finch/tensor.py
```diff
@@ -587,25 +615,38 @@ def asarray(obj, /, *, dtype=None, format=None):
         return tensor
 
 
+def reshape(x: Tensor, /, shape: tuple[int, ...], *, copy: bool | None = None) -> Tensor:
+    # TODO: https://github.com/willow-ahrens/Finch.jl/issues/558
+    #       Only to run array-api-tests that require it for multiple tests.
+    #       Must be reimplemented once `reshape` is available in Finch.jl.
+    arr = x.todense()
```
willow-ahrens, line 619: Could we simply add a PerformanceWarning here that we are densifying?

willow-ahrens (changes requested):
Let's just add a warning for the temporary reshape. Otherwise LGTM!

src/finch/tensor.py
hameerabbasi: I would add `format` to `compute`, to minimise the amount of additional API added. Uniformity is good, less API to do the same thing is even better.

willow-ahrens: I think it's very reasonable for a user to want an empty COO matrix full of ones.

willow-ahrens: so they would type `full(1, format=coo)`. You're proposing that they need to do `compute(lazy(full(1)), format=coo)`?

willow-ahrens: or would it be `compute(full(lazy(1)), format="coo")`? I'm not quite sure where the lazy would go in this invocation. Another issue is that we may want to specify intermediate formats for performance reasons, so it's not just the final format that matters. Finally, it would be nice to be able to change the format without dropping into the lazy interface.

hameerabbasi: I think it would be close to `full(1).compute(format[, optional_schedule])` in a lazy context (i.e. inside a function with `@compiled`.

`compute` would force materialisation in any context.

I guess a middle way would be to add it to creation functions only, but every function would be way too much API surface.

willow-ahrens: I'm not sure we can support a lazy version of this function. I agree that adding `format` to all creation functions is a good plan.

src/finch/tensor.py
mtsokol: Done!

mtsokol:
Hi @willow-ahrens @hameerabbasi Can we merge this PR? Do you have any other comments?

willow-ahrens (approved):
Lgtm!

src/finch/dtypes.py
```diff
@@ -40,6 +43,17 @@
     float64: np.float64,
     complex64: np.complex64,
     complex128: np.complex128,
-    bool: np.bool_,
+    bool: builtins.bool,
```
hameerabbasi, line 46: This is weird, all the others here are NumPy dtypes.

src/finch/tensor.py
```diff
@@ -86,7 +87,7 @@ def __init__(
         *,
         fill_value: np.number = 0.0,
     ):
-        if isinstance(obj, (int, float, complex, bool)):
+        if isinstance(obj, (int, float, complex, bool, list)):
```
hameerabbasi, line 90: I'd avoid hard-coding list and check for `Iterable`.

src/finch/tensor.py
```diff
@@ -897,8 +949,32 @@ def atanh(x: Tensor, /) -> Tensor:
...
+def logical_and(x1: Tensor, x2: Tensor, /) -> Tensor:
+    return x1._elemwise_op("Finch.and", x2)
+
+
+def logical_or(x1: Tensor, x2: Tensor, /) -> Tensor:
+    return x1._elemwise_op("Finch.or", x2)
+
+
+def logical_xor(x1: Tensor, x2: Tensor, /) -> Tensor:
+    return x1._elemwise_op("Finch.xor", x2)
```
hameerabbasi, line 977: Can we add tests for these?

hameerabbasi (changes requested):
A few comments to finish up, seems like the larger discussions are resolved.

src/finch/tensor.py
mtsokol: They are all tested. Please see `test_elemwise_ops_2_args` and `test_elemwise_ops_1_arg` changes.

src/finch/dtypes.py
mtsokol: The same change already happens in `tensor.py::638`:
```
if dtype == np.bool_:  # Fails with: Finch currently only supports isbits defaults
    dtype = bool
```
`np.bool_` isn't accepted by Finch, only builtin `bool` is fine.

hameerabbasi: Is this a `Finch.jl` issue or a Julia issue, @willow-ahrens?

src/finch/tensor.py
mtsokol: But that would be incorrect. We only support iterables that can be passed to `np.array(...)` to create a valid array.

`dict` and `set` are out of support (only `list` and `tuple`):
```python
In [9]: isinstance(set((1,2,3)), Iterable)
Out[9]: True

In [10]: np.array(set((1,2,3)))
Out[10]: array({1, 2, 3}, dtype=object)

In [12]: isinstance({1:2}, Iterable)
Out[12]: True

In [13]: np.array({1:2})
Out[13]: array({1: 2}, dtype=object)
```

src/finch/dtypes.py
mtsokol: @hameerabbasi Can we inspect it in a follow-up work? We already use `np.bool_` -> `builtins.bool` in a few places in `tensor.py`, I don't think it should be addressed here.

hameerabbasi: Sure, I'll create an issue.

hameerabbasi (approved):

PR #50, opened 2024-05-14
add floor division

willow-ahrens:
This is a PR to add floor division based on https://github.com/willow-ahrens/Finch.jl/pull/557


mtsokol:
Thank you for a PR! `__floordiv__` already exists in `Tensor`, you need to rebase as you're 65 commits behind.

willow-ahrens:
oops! I just made a different PR to keep things simple.

PR #51, opened 2024-05-14
fix floordiv and mod

willow-ahrens:
Uses the new `fld_nothrow` and `rem_nothrow` functions in Finch to do floor division and remainder on sparse arrays with zeros without erroring.


mtsokol (approved):
LGTM! I enabled tests. I also added `__abs__` to `__mod__` function because in Python:
```
In [75]: -10 % 15
Out[75]: 5
```
where in Julia:
```
julia> -10 % 15
-10
```

hameerabbasi:
Ouch -- `__abs__` was a mistake, imho. We should add the divisor instead if the result is negative.

willow-ahrens:
got it, there's a fix on its way

PR #55, opened 2024-05-17
Make `nonzero` return Tensors

mtsokol:
Hi @hameerabbasi,

This is a small tweak to make sure `nonzero` returns Tensor's instead of NumPy arrays.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "finch-tensor"
-version = "0.1.20"
+version = "0.1.21"
 description = ""
 authors = ["Willow Ahrens <willow.marie.ahrens@gmail.com>"]
 readme = "README.md"
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -723,7 +723,7 @@ def nonzero(x: Tensor, /) -> tuple[np.ndarray, ...]:
     indices = jl.ffindnz(x._obj)[:-1]  # return only indices, skip values
     indices = tuple(np.asarray(i) - 1 for i in indices)
     sort_order = np.lexsort(indices[::-1])  # sort to row-major, C-style order
-    return tuple(i[sort_order] for i in indices)
+    return tuple(Tensor(i[sort_order]) for i in indices)
 
 
 def _reduce(x: Tensor, fn: Callable, axis, dtype=None):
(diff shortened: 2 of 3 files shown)


hameerabbasi (approved):

willow-ahrens:
Is this the desired approach? I'm fine with this but I wonder if users would want a wrapped array when they call these destructuring operators. Usually they call this when the arrays are moving to a different library or framework.

willow-ahrens:
Hameer, you've been right before when it comes to conventions like this, so I'll defer to you here. This struck me as strange though since most of the time people call this when they "just want the data"

hameerabbasi:
> Hameer, you've been right before when it comes to conventions like this, so I'll defer to you here. This struck me as strange though since most of the time people call this when they "just want the data"

Well, what I'd do is compute `arr == zero_scalar` as a `Tensor` with a dtype of `bool`, materialize it as a canonicalized `COO`, and pull out the `coords`.

mtsokol:
I wrapped it because `array-api-tests` expects results of functions of the same type as inputs. So, Tensor in, Tensor out. When Tensor is an input and NumPy is an output it flags it as an error.

hameerabbasi:
> I wrapped it because `array-api-tests` expects results of functions of the same type as inputs. So, Tensor in, Tensor out. When Tensor is an input and NumPy is an output it flags it as an error.

Yes, we should definitely have "this library type" in/out.

mtsokol:
> > Hameer, you've been right before when it comes to conventions like this, so I'll defer to you here. This struck me as strange though since most of the time people call this when they "just want the data"
> 
> Well, what I'd do is compute `arr == zero_scalar` as a `Tensor` with a dtype of `bool`, materialize it as a canonicalized `COO`, and pull out the `coords`.

We already have `nonzero` implementation on the Finch side, which we're calling here.

willow-ahrens:
Ah, I see. No worries, we can leave as is if the Array API requires it. Just thought I'd register that this is usually not what the user would expect from a function like this. A related function is "find". In Scipy, "find" does not return a tuple of `scipy.sparse` arrays, it just returns ndarray. https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.find.html#scipy.sparse.find

willow-ahrens:
I think most people will find themselves unwrapping the results of this function

willow-ahrens:
Maybe if `find` is not in the API standard, we could introduce a function called `find` that just returns ndarrays.

PR #56, opened 2024-05-17
fix mod

### LINKED ISSUE #53: `__mod__` can return incorrect results
Ouch -- `__abs__` was a mistake, imho. We should add the divisor instead if the result is negative.

_Originally posted by @hameerabbasi in https://github.com/willow-ahrens/finch-tensor/issues/51#issuecomment-2116987378_

willow-ahrens:
fixes https://github.com/willow-ahrens/finch-tensor/issues/53

### DIFF
--- a/src/finch/julia.py
+++ b/src/finch/julia.py
@@ -3,7 +3,7 @@ import os
 import juliapkg
 
 _FINCH_NAME = "Finch"
-_FINCH_VERSION = "0.6.27"
+_FINCH_VERSION = "0.6.28"
 _FINCH_HASH = "9177782c-1635-4eb9-9bfb-d9dfa25e6bce"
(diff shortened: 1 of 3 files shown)


hameerabbasi (approved):
Thanks for the lightning quick fix, @willow-ahrens!

hameerabbasi:
Minor comment: Could we add a regression test?

willow-ahrens:
Is this not tested in the array api tests?

willow-ahrens:
I'll be fairly busy the next couple of weeks, so I'd appreciate help on these tests.

mtsokol (approved):

mtsokol:
I added a regression test. Thanks @willow-ahrens!

PR #60, opened 2024-05-27
Add `device` keyword to creation functions

mtsokol:
Hi @hameerabbasi,

This PR adds `device` keyword to creation functions to conform with Array API (Only accepts `"cpu"` now).

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -18,7 +18,7 @@ from .levels import (
     SparseList,
     sparse_formats_names,
 )
-from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays, DType
+from .typing import OrderType, JuliaObj, spmatrix, TupleOf3Arrays, DType, Device
 
 
 class SparseArray:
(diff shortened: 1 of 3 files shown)


hameerabbasi (changes requested):
Thanks, LGTM besides the one comment.

src/finch/tensor.py
```diff
@@ -683,34 +686,53 @@ def full_like(
     *,
     dtype: DType | None = None,
     format: str = "coo",
+    device: Literal["cpu"] | None = None,
```
```suggestion
    device: Device = None,
```

And let's define `Device` in `finch.typing`?

hameerabbasi (approved):

willow-ahrens:
Finch does have a notion of device, though it's not quite ready for production. You can view it in https://github.com/willow-ahrens/Finch.jl/blob/main/src/architecture.jl

willow-ahrens:
Maybe once that's up and running we can port it to python somehow, but I think we'll need to discuss some more.

PR #61, opened 2024-05-28
Add `real`, `imag`, and `conj` functions

mtsokol:
Hi @hameerabbasi,

This PR adds complex functions: `real`, `imag`, and `conj`

### DIFF
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -77,6 +77,19 @@ def test_elemwise_ops_1_arg(arr3d, func_name):
 
 
 @pytest.mark.parametrize(
+    "func_name", ["real", "imag", "conj"]
+)
+def test_elemwise_complex_ops_1_arg(func_name):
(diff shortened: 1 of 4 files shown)


tests/test_ops.py
```diff
@@ -76,6 +76,19 @@ def test_elemwise_ops_1_arg(arr3d, func_name):
     assert_allclose(actual.todense(), expected)
 
 
+@pytest.mark.parametrize(
+    "func_name", ["real", "imag", "conj"]
+)
+def test_elemwise_complex_ops_1_arg(func_name):
```
hameerabbasi: Could we test more dtypes? I don't know if Finch would do the right thing for those.

hameerabbasi (changes requested):
One comment.

tests/test_ops.py
mtsokol: Done!

hameerabbasi (approved):

PR #64, opened 2024-06-05
Refine `dtype` argument for `sum` and `prod`

mtsokol:
Hi @hameerabbasi @willow-ahrens,

This PR adds proper support for the `dtype` argument for `finch.sum` and `finch.prod`.

It follows [Array API standard rules](https://data-apis.org/array-api/latest/API_specification/generated/array_api.sum.html), specifically for integer dtypes (there's a promotion rule saying that for `dtype=None` signed and unsigned dtypes should be promoted to the default (signed or unsigned) integer).

`_reduce` also makes sure that scalars returned from Julia (when `axis=None`, so it sums to a scalar) have the correct dtype. This turned out to be a bit tricky. If we're summing `Int8` Tensor the scalar result is returned as a Python builtin (so e.g. int64) so dtype is "erased" when Julia returns the result of e.g. `jl.sum`. To recreate Finch Tensor 0D array/scalar with the correct dtype I did:
```python
if np.isscalar(result):
    result = jl.Tensor(
        jl.Element(
            jc.convert(x.dtype, 0),
            np.array(result, dtype=jl_dtypes.jl_to_np_dtype[x.dtype])
        )
    )
```
Which is a bit cumbersome but it works.

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -743,14 +743,14 @@ def astype(x: Tensor, dtype: DType, /, *, copy: bool = True):
     if not copy:
         if x.dtype == dtype:
             return x
-        else:
+        if copy is False:
             raise ValueError("Unable to avoid a copy while casting in no-copy mode.")
-    else:
-        finch_tns = x._obj.body
-        result = jl.copyto_b(
-            jl.similar(finch_tns, jc.convert(dtype, jl.default(finch_tns)), dtype), finch_tns
-        )
-        return Tensor(jl.swizzle(result, *x.get_order(zero_indexing=False)))
+
+    finch_tns = x._obj.body
+    result = jl.copyto_b(
+        jl.similar(finch_tns, jc.convert(dtype, jl.default(finch_tns)), dtype), finch_tns
+    )
+    return Tensor(jl.swizzle(result, *x.get_order(zero_indexing=False)))
 
 
 def where(condition: Tensor, x1: Tensor, x2: Tensor, /) -> Tensor:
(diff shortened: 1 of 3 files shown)


src/finch/tensor.py
```diff
@@ -785,7 +785,25 @@ def _reduce(x: Tensor, fn: Callable, axis, dtype=None):
...
+    if np.isscalar(result):
+        result = jl.Tensor(
+            jl.Element(
+                jc.convert(x.dtype, 0),
+                np.array(result, dtype=jl_dtypes.jl_to_np_dtype[x.dtype])
+            )
+        )
+
+    result = Tensor(result)
+    if dtype != "skip":
```
hameerabbasi: What purpose does `dtype="skip"` serve? I don't think we should ever return Python scalars.

src/finch/tensor.py
```diff
@@ -785,7 +785,25 @@ def _reduce(x: Tensor, fn: Callable, axis, dtype=None):
...
+        # dtype casting rules for `sum` and `prod`
+        if dtype is not None:
+            result = astype(result, dtype)
+        elif jl.seval(f"{x.dtype} <: Unsigned"):
+            result = astype(result, jl_dtypes.uint)
+        elif jl.seval(f"{x.dtype} <: Signed"):
+            result = astype(result, jl_dtypes.int_)
```
hameerabbasi: The dtype promotion rules are invalid in the 2023 version of the Array API standard -- we shouldn't write code to cater to an old version of the standard.

hameerabbasi (changes requested):
One question and a concern.

src/finch/tensor.py
hameerabbasi: Ah, that's my bad, it does exist in 2023.

src/finch/tensor.py
hameerabbasi: Can we flip this logic, so that the default behaviour is the common one.

src/finch/tensor.py
mtsokol: I think they are still valid in `2023` version: https://data-apis.org/array-api/2023.12/API_specification/generated/array_api.sum.html

> if x has a signed integer data type (e.g., int16), the returned array must have the default integer data type.

> if x has an unsigned integer data type (e.g., uint16), the returned array must have an unsigned integer data type having the same number of bits as the default integer data type (e.g., if the default integer data type is int32, the returned array must have a uint32 data type).

src/finch/tensor.py
mtsokol: These `dtype` rules only apply to `sum` and `prod`. The other reduction functions, `max` etc., skip it.

hameerabbasi (approved):
Thanks for the excellent work here @mtsokol!

PR #70, opened 2024-06-13
Port `read` and `write` functions

mtsokol:
 
Here's a minimal PR to port `fread` and `fwrite` to `finch-tensor`.

### DIFF
--- a/src/finch/io.py
+++ b/src/finch/io.py
@@ -1,11 +1,13 @@
+from pathlib import Path
+
 from .julia import jl
 from .tensor import Tensor
 
 
-def read(filename: str) -> Tensor:
-    julia_obj = jl.fread(filename)
+def read(filename: Path | str) -> Tensor:
+    julia_obj = jl.fread(str(filename))
     return Tensor(julia_obj)
 
 
-def write(filename: str, tns: Tensor) -> None:
-    jl.fwrite(filename, tns._obj)
+def write(filename: Path | str, tns: Tensor) -> None:
+    jl.fwrite(str(filename), tns._obj)
--- a/src/finch/io.py
+++ b/src/finch/io.py
@@ -1,13 +1,35 @@
 from pathlib import Path
 
-from .julia import jl
+from .julia import add_package, jl
 from .tensor import Tensor
 
 
+def _import_deps(filename: str) -> None:
+    fn = filename
+    if fn.endswith(".mtx") or fn.endswith(".ttx") or fn.endswith(".tns"):
+        add_package("TensorMarket", hash="8b7d4fe7-0b45-4d0d-9dd8-5cc9b23b4b77", version="0.2.0")
+        jl.seval("using TensorMarket")
+    elif fn.endswith(".bspnpy"):
+        add_package("NPZ", hash="15e1cf62-19b3-5cfa-8e77-841668bca605", version="0.4.3")
+        jl.seval("using NPZ")
+    elif fn.endswith(".bsp.h5"):
+        add_package("HDF5", hash="f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f", version="0.17.2")
+        jl.seval("using HDF5")
+    else:
+        raise ValueError(
+            f"Unsupported file extension. Supported extensions are "
+            "`.mtx`, `.ttx`, `.tns`, `.bspnpy`, and `.bsp.h5`."
+        )
+
+
 def read(filename: Path | str) -> Tensor:
-    julia_obj = jl.fread(str(filename))
+    fn = str(filename)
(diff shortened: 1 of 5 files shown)


tests/test_io.py
```diff
@@ -0,0 +1,25 @@
...
+
+    assert_equal(tns.todense(), arr2d)
+
+
+def test_write(arr2d):
+    tns = finch.asarray(arr2d)
+    finch.write(f"{base_path}/tmp.ttx", tns)
+
+    expected = open(f"{base_path}/matrix_1.ttx").read()
+    actual = open(f"{base_path}/tmp.ttx").read()
```
hameerabbasi: I'd use the `tempfile` module here.

hameerabbasi: Or, even better: https://docs.pytest.org/en/latest/how-to/tmp_path.html

src/finch/io.py
```diff
@@ -0,0 +1,11 @@
+from .julia import jl
+from .tensor import Tensor
+
+
+def read(filename: str) -> Tensor:
+    julia_obj = jl.fread(filename)
```
hameerabbasi: It'd be nice to accept `pathlib.Path` here too.

src/finch/io.py
```diff
@@ -0,0 +1,11 @@
...
+    return Tensor(julia_obj)
+
+
+def write(filename: str, tns: Tensor) -> None:
+    jl.fwrite(filename, tns._obj)
```
hameerabbasi: Same here, `pathlib.Path` support would be nice.

src/finch/julia.py
```diff
@@ -30,3 +30,4 @@
 
 jl.seval("using Finch")
 jl.seval("using Random")
+jl.seval("using TensorMarket")
```
hameerabbasi: Will this limit us to just this format?

mtsokol: No, it's just a dependency. Here's a list of all supported formats: https://willowahrens.io/Finch.jl/stable/guides/fileio/#Finch.fread

hameerabbasi: Do we need to pull in any other dependencies for other formats?

src/finch/julia.py
```diff
@@ -8,20 +8,30 @@
...
-        juliapkg.add(_FINCH_NAME, _FINCH_HASH, version=_FINCH_VERSION)
+elif (
+    deps.get("packages", {}).get(_FINCH_NAME, {}).get("version", None)
+    != _FINCH_VERSION
+):
+    juliapkg.add(_FINCH_NAME, _FINCH_HASH, version=_FINCH_VERSION)
+
+if (
+    deps.get("packages", {}).get(_TENSOR_MARKET_NAME, {}).get("version", None)
+    != _FINCH_VERSION
```
```suggestion
    != _TENSOR_MARKET_VERSION
```

Also, since we're now doing this logic twice, maybe we can pull it into a function and call that function twice? That way we can also get rid of the extra global vars.

tests/test_io.py
mtsokol: Done!

src/finch/io.py
mtsokol: Done!

src/finch/io.py
mtsokol: Done!

src/finch/julia.py
mtsokol: @willow-ahrens Are there other IO related dependencies that we should install during the setup to fully support `fread` and `fwrite`?

In `Finch.jl` in `Project.toml` I found three more:

```toml
[weakdeps]
HDF5 = "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
NPZ = "15e1cf62-19b3-5cfa-8e77-841668bca605"
SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
TensorMarket = "8b7d4fe7-0b45-4d0d-9dd8-5cc9b23b4b77"
```

willow-ahrens: HDF5, NPZ, and TensorMarket are required for binsparse and ttx files.

willow-ahrens: These should be added with pkg.add when we add Finch, or loaded dynamically depending on whether we want to take a dependency on hdf5

mtsokol: Thanks! I load them dynamically.

src/finch/julia.py
mtsokol: Done!

mtsokol:
@hameerabbasi it's ready for another review. All IO dependencies are loaded dynamically (once) when a specific file extension is present.

hameerabbasi (approved):

PR #72, opened 2024-06-17
Add `empty`, `empty_like`, `arange`, and `linspace`

mtsokol:
Hi @hameerabbasi,

This is the last PR with the creation functions. `empty` and `empty_like` are implemented the same way as in Numba backend. `arange` and `linspase` are strictly dense functions therefore I decided to use NumPy functions and pass results to Tensor (no-copy operation).

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -785,6 +785,57 @@ def zeros_like(
     return zeros(x.shape, dtype=dtype, format=format, device=device)
 
 
+def empty(
+    shape: int | tuple[int, ...],
+    *,
+    dtype: DType | None = None,
+    format: str = "coo",
+    device: Device = None,
+) -> Tensor:
+    return full(shape, np.float64(0), dtype=dtype, format=format, device=device)
+
+
+def empty_like(
+    x: Tensor,
+    /,
+    *,
+    dtype: DType | None = None,
+    format: str = "coo",
+    device: Device = None,
+) -> Tensor:
+    dtype = x.dtype if dtype is None else dtype
+    return empty(x.shape, dtype=dtype, format=format, device=device)
+
+
+def arange(
+    start: int | float,
+    /,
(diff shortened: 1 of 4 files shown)


src/finch/tensor.py
```diff
@@ -785,6 +785,57 @@ def zeros_like(
...
+    num: int,
+    *,
+    dtype: DType | None = None,
+    device: Device = None,
+    endpoint: bool = True,
+) -> Tensor:
+    _validate_device(device)
+    return Tensor(
+        np.linspace(start, stop, num=num, dtype=jl_dtypes.jl_to_np_dtype[dtype], endpoint=endpoint)
+    )
```
hameerabbasi, line 836: We should add minimal tests for these two.

hameerabbasi (approved):
Do the tests now pass in the Array API for these functions?

mtsokol:
> Do the tests now pass in the Array API for these functions?

Yes! Here's the diff:
```diff
--- a/ci/Finch-array-api-xfails.txt
+++ b/ci/Finch-array-api-xfails.txt
@@ -72,10 +72,6 @@ array_api_tests/test_signatures.py::test_array_method_signature[__setitem__]
 # test_creation_functions

 # not implemented
-array_api_tests/test_creation_functions.py::test_arange
-array_api_tests/test_creation_functions.py::test_empty
-array_api_tests/test_creation_functions.py::test_empty_like
-array_api_tests/test_creation_functions.py::test_linspace
 array_api_tests/test_creation_functions.py::test_meshgrid

 # test_array_object
```

hameerabbasi:
From the CI failures, looks like we need a NumPy 2.0 compat PR. 🤔

src/finch/tensor.py
mtsokol: Done!

mtsokol:
It was a typo in the test change. Rebased.

PR #75, opened 2024-06-24
 MAINT: Refine order support

### LINKED ISSUE #74: `asarray` should convert format if necessary
> Ok, so we have this code snippet from the notebook that uses Finch backend:
> 
> ```python
> X = sparse.random((100, 5), density=0.08)  # creates COO random matrix
> X = sparse.asarray(X, format="csc")  # converts to CSC format
> X_X = sparse.permute_dims(X, (1, 0)) @ X  # for me locally it densifies as the result is: SwizzleArray(Tensor(Dense{Int64}(Dense{Int64}(Element{0.0, Float64, Int64}...
> 
> X_X.get_order()  # it gives (1, 0) order so I can only convert to CSR. 
> 
> X_X = sparse.asarray(X_X, format="csr")  # move back from dense to CSR format
> ```
> 
> So it looks like in the CI the result order of `permute_dims(X, (1, 0)) @ X` is sometimes `(0, 1)` as it's the only reason for this failure.

While the indeterminism here is definitely an issue, one other issue is that `asarray` should convert the format if necessary, similar to the `order=` or `dtype=` kwargs.

_Originally posted by @hameerabbasi in https://github.com/pydata/sparse/issues/709#issuecomment-2186224847_

mtsokol:

This PR refines order support, by allowing `finch.asarray(any_input, format=any_format)` regardless of `any_input` order.

Right now it's accomplished by calling `copyto!` and then `dropfills` which I think performs copy twice. Could it be achieved with a one function call? (or performing one copy)

- `copyto!` for dense source and sparse (e.g. `csr`) destination copies all elements including fill values as non-fill ones (therefore it requires `dropfills`)
- calling only `dropfills!` fails for `csr` format as shown here https://github.com/willow-ahrens/Finch.jl/issues/609#issuecomment-2186636101

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -388,7 +388,7 @@ class Tensor(_Display, SparseArray):
         result = jl.copyto_b(
             jl.swizzle(jl.Tensor(storage.levels_descr._obj), *order), tensor._obj
         )
-        return jl.dropfills(result)
+        return jl.dropfills(result) if tensor._is_dense else result
 
     @classmethod
     def _from_numpy(cls, arr: np.ndarray, fill_value: np.number, copy: bool | None = None) -> JuliaObj:
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -383,11 +383,12 @@ class Tensor(_Display, SparseArray):
         return Tensor(self._from_other_tensor(self, storage=storage))
 
     @classmethod
-    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage | None) -> JuliaObj:
+    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage) -> JuliaObj:
         order = cls.preprocess_order(storage.order, tensor.ndim)
-        return jl.swizzle(
-            jl.Tensor(storage.levels_descr._obj, tensor._obj.body), *order
+        result = jl.copyto_b(
+            jl.swizzle(jl.Tensor(storage.levels_descr._obj), *order), tensor._obj
         )
+        return jl.dropfills(result)
 
     @classmethod
     def _from_numpy(cls, arr: np.ndarray, fill_value: np.number, copy: bool | None = None) -> JuliaObj:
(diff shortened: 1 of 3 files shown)


src/finch/tensor.py
```diff
@@ -383,11 +383,12 @@ def to_storage(self, storage: Storage) -> "Tensor":
...
     @classmethod
-    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage | None) -> JuliaObj:
+    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage) -> JuliaObj:
         order = cls.preprocess_order(storage.order, tensor.ndim)
-        return jl.swizzle(
-            jl.Tensor(storage.levels_descr._obj, tensor._obj.body), *order
+        result = jl.copyto_b(
+            jl.swizzle(jl.Tensor(storage.levels_descr._obj), *order), tensor._obj
         )
+        return jl.dropfills(result)
```
hameerabbasi: I'd remove the `jl.dropfills` call unless `tensor` is dense, which is really the only case where it matters much.

src/finch/tensor.py
```diff
@@ -383,11 +383,12 @@ def to_storage(self, storage: Storage) -> "Tensor":
...
     @classmethod
-    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage | None) -> JuliaObj:
+    def _from_other_tensor(cls, tensor: "Tensor", storage: Storage) -> JuliaObj:
         order = cls.preprocess_order(storage.order, tensor.ndim)
-        return jl.swizzle(
-            jl.Tensor(storage.levels_descr._obj, tensor._obj.body), *order
+        result = jl.copyto_b(
+            jl.swizzle(jl.Tensor(storage.levels_descr._obj), *order), tensor._obj
         )
+        return jl.dropfills(result) if tensor._is_dense else result
```
hameerabbasi, line 391: Does this always copy? We should avoid a copy if at all possible.

mtsokol: Are you asking if `dropfills` always copy?

hameerabbasi: I'm asking if `copyto_b` always copies.

mtsokol: Yes, I think `copyto_b` always copies.

mtsokol: I think `copyto_b` is used here correctly. It's here to implement `to_storage` that is meant to copy Tensor to a new Storage.

hameerabbasi: Is there a way to avoid the copy in some cases, e.g. when the format is the same?

mtsokol: I can include it in `asarray` later, once we can compare tensors formats. Right now there's no way to know if a given tensor is `csr` or `csc` so we don't know if the `format` argument is the same.

hameerabbasi: What I'm worried about is `xp.asarray` always copying.

hameerabbasi (approved):

src/finch/tensor.py
mtsokol: `asarray` is always copying when: `format` is provided (as explained in my previous comment), `copy=True`, or `copy=None` and `dtype` is different from the current one. Otherwise Tensor is created in a no-copy manner.

PR #82, opened 2024-12-06
Update `complied` function

mtsokol:
Hi @hameerabbasi,

One missing thing from yesterday's work is `def complied` update, i.e. removal of `get_scheduler(...)` call.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "finch-tensor"
-version = "0.2.0"
+version = "0.2.1"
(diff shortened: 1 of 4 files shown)


hameerabbasi (commented):
Can we add tests showing how this would be used?

mtsokol:
> Can we add tests showing how this would be used?

Sure! I tweaked `test_lazy_mode` test so that `opt` param is actually passed there.

hameerabbasi (approved):
Feel free to merge once CI is green!

PR #83, opened 2024-12-12
Allow for recursive `@compiled` without materialization

hameerabbasi:
- **Materialize only in outermost nested `@compiled`.**
- **`pre-commit autoupdate`**

### DIFF
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -4,20 +4,53 @@ from .julia import jl
 from .tensor import Tensor
 
 
-def compiled(opt=None):
+def _recurse(x, /, *, f):
+    if isinstance(x, tuple | list):
+        return type(x)(_recurse(xi) for xi in x)
+    if isinstance(x, dict):
+        return {k: _recurse(v) for k, v in x}
+    return f(x)
+
+
+def _recurse_iter(x, /):
+    if isinstance(x, tuple | list):
+        yield from (_recurse(xi) for xi in x)
+    if isinstance(x, dict):
+        yield from (_recurse(xi) for xi in x.values())
+    yield x
+
+
+def _to_lazy_tensor(x, /):
+    if isinstance(x, Tensor) and not jl.isa(x._obj, jl.Finch.LazyTensor):
+        return Tensor(jl.Finch.LazyTensor(x._obj))
+
+    return x
+
+
+def compiled(opt=None, force_materialization=False):
     def inner(func):
         @wraps(func)
         def wrapper_func(*args, **kwargs):
-            new_args = []
-            for arg in args:
-                if isinstance(arg, Tensor) and not jl.isa(arg._obj, jl.Finch.LazyTensor):
-                    new_args.append(Tensor(jl.Finch.LazyTensor(arg._obj)))
-                else:  
-                    new_args.append(arg)
-            result = func(*new_args, **kwargs)
-            kwargs = {"ctx": opt.get_julia_scheduler()} if opt is not None else {}
-            result_tensor = Tensor(jl.Finch.compute(result._obj, **kwargs))
+            args = tuple(args)
+            kwargs = dict(kwargs)
+            compute_at_end = True
+            if not force_materialization and any(
+                isinstance(arg, Tensor) and jl.isa(arg._obj, jl.Finch.LazyTensor)
+                for arg in _recurse_iter((args, kwargs))
+            ):
+                compute_at_end = False
+
+            args = _recurse(args, f=_to_lazy_tensor)
+            kwargs = _recurse(kwargs, f=_to_lazy_tensor)
+            result = func(*args, **kwargs)
+            if not compute_at_end:
+                return result
+            compute_kwargs = (
+                {"ctx": opt.get_julia_scheduler()} if opt is not None else {}
+            )
+            result_tensor = Tensor(jl.Finch.compute(result._obj, **compute_kwargs))
             return result_tensor
+
         return wrapper_func
 
     return inner
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -6,17 +6,17 @@ from .tensor import Tensor
 
 def _recurse(x, /, *, f):
     if isinstance(x, tuple | list):
-        return type(x)(_recurse(xi) for xi in x)
+        return type(x)(_recurse(xi, f=f) for xi in x)
     if isinstance(x, dict):
-        return {k: _recurse(v) for k, v in x}
+        return {k: _recurse(v, f=f) for k, v in x}
     return f(x)
 
 
 def _recurse_iter(x, /):
     if isinstance(x, tuple | list):
-        yield from (_recurse(xi) for xi in x)
+        yield from (_recurse_iter(xi) for xi in x)
     if isinstance(x, dict):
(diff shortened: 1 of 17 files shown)


src/finch/tensor.py
```diff
@@ -154,14 +157,27 @@ def __mod__(self, other):
...
-        if self.ndim != 2 or other.ndim != 2:
-            raise ValueError(
-                f"Both tensors must be 2-dimensional, but are: {self.ndim=} and {other.ndim=}."
+    @compiled()
+    def __matmul__(self, other: "Tensor") -> "Tensor":
+        if self.ndim < 2 or other.ndim < 2:
+            raise NotImplementedError(
+                "`@` is not implemented for arrays with less than two dimensions."
             )
-        return tensordot(self, other, axes=((-1,), (-2,)))
```
mtsokol, line 164: As I checked today, with Finch 1.0 we don't need this hack - `tensordot` is fine to use.

mtsokol:
I'll review it once it's rebased.

src/finch/tensor.py
hameerabbasi: This is more general as well; supports tensors with ndim >= 2.

mtsokol: I think we still want this change reverted.

hameerabbasi: Do we know if tensordot supports ndim > 2? If not I’d prefer to keep it this way.

willow-ahrens: tensordot supports ndim > 2

hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/83/commits/1f6513c7da4b1516c34e75484e912338eaa7c7b6.

src/finch/tensor.py
```diff
@@ -154,12 +157,11 @@ def __mod__(self, other):
     def __pow__(self, other):
         return self._elemwise_op("^", other)
 
-    def __matmul__(self, other):
-        # TODO: Implement and use mul instead of tensordot
-        # https://github.com/FinchTensor/finch-tensor-python/pull/22#issuecomment-2007884763
-        if self.ndim != 2 or other.ndim != 2:
-            raise ValueError(
-                f"Both tensors must be 2-dimensional, but are: {self.ndim=} and {other.ndim=}."
+    @compiled()
```
mtsokol: We don't need `@compiled` decorator here.

src/finch/tensor.py
```diff
@@ -18,6 +18,7 @@
     SparseList,
     sparse_formats_names,
 )
+from .compiled import compiled
```
mtsokol: No need to import it.

src/finch/compiled.py
```diff
@@ -2,10 +2,66 @@
 from functools import wraps
 
 from .julia import jl
-from .tensor import Tensor
```
mtsokol, line 5: We can't import it top level anymore?

src/finch/compiled.py
```diff
@@ -2,10 +2,66 @@
...
+def _to_lazy_tensor(x, /):
+    from .tensor import Tensor
+
+    if isinstance(x, Tensor) and not jl.isa(x._obj, jl.Finch.LazyTensor):
+        return Tensor(jl.Finch.LazyTensor(x._obj))
+
+    return x
+
+
+def compiled(opt=None, *, force_materialization=False, tag: int | None = None):
```
mtsokol, line 32: As it's a new feature I would like to see a few tests for it:
- Test with a nested `@compiled` decorators.
- Test with `force_materialization=True` which checks that for all lazy inputs the output is computed.

mtsokol (changes requested):

src/finch/compiled.py
hameerabbasi: It led to a recursive import; anyway it's undone in https://github.com/finch-tensor/finch-tensor-python/pull/83/commits/87f384c06d507d9c9024d880b347977fa7a14a5f.

src/finch/tensor.py
hameerabbasi: Fixed in https://github.com/finch-tensor/finch-tensor-python/pull/83/commits/87f384c06d507d9c9024d880b347977fa7a14a5f.

src/finch/compiled.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/83/commits/4d45756fa3ed78c2c8785a15c81ffeda94d5cac8.

src/finch/tensor.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/83/commits/87f384c06d507d9c9024d880b347977fa7a14a5f.

src/finch/compiled.py
```diff
@@ -2,8 +2,58 @@
 from functools import wraps
 
 from .julia import jl
+from .typing import JuliaObj, Any
 from .tensor import Tensor
-from .typing import JuliaObj
+
+
+def _recurse(x, /, *, f):
```
mtsokol: Let's put type annotations back in.

src/finch/compiled.py
```diff
@@ -2,8 +2,58 @@
...
+    if isinstance(x, tuple | list):
+        return type(x)(_recurse(xi, f=f) for xi in x)
+    if isinstance(x, dict):
+        return {k: _recurse(v, f=f) for k, v in x}
+    return f(x)
+
+
+def _recurse_iter(x, /):
```
mtsokol: Let's put type annotations back in.

src/finch/compiled.py
```diff
@@ -2,8 +2,58 @@
...
+    if isinstance(x, tuple | list):
+        for xi in x:
+            yield from _recurse_iter(xi)
+    if isinstance(x, dict):
+        for xi in x.values():
+            yield from _recurse_iter(xi)
+    yield x
+
+
+def _to_lazy_tensor(x: Tensor | Any, /):
```
mtsokol: Return type.

src/finch/compiled.py
```diff
@@ -29,46 +79,13 @@ def set_optimizer(opt: AbstractScheduler) -> None:
...
-            result_tensor = Tensor(jl.Finch.compute(result._obj, **kwargs))
-            return result_tensor
-
-        return wrapper_func
-
-    return inner
-
-
-def lazy(tensor: Tensor) -> Tensor:
+def lazy(tensor: Tensor):
```
mtsokol: Return type.

src/finch/compiled.py
```diff
@@ -29,46 +79,13 @@ def set_optimizer(opt: AbstractScheduler) -> None:
...
     if tensor.is_computed():
         return Tensor(jl.Finch.LazyTensor(tensor._obj))
     return tensor
 
 
-def compute(
-    tensor: Tensor, *, opt: AbstractScheduler | None = None, tag: int = -1
-) -> Tensor:
+def compute(tensor: Tensor, *, opt: AbstractScheduler | None = None, tag: int = -1):
```
mtsokol: Return type.

mtsokol (approved):
LGTM! Just typing nitpicks and it's ready.

PR #84, opened 2024-12-12
propagate the `@compiled` tag parameter to finch

mtsokol:
This PR propagates `tag` parameter and refactors `compiled.py`.


hameerabbasi (approved):
LGTM; thanks for this @mtsokol

PR #88, opened 2025-02-03
(fix): checking `None` membership in `__getitem__`

ilan-gold:
I ran into this issue when trying to index ~with a tuple~ that had a numpy array in it:

```python
import finch
import numpy as np

tensor = finch.Tensor(np.array([[1, 2], [3, 4]]))
tensor[finch.Tensor(np.array([False, True]))]
```

Previously, `None in key` would trigger `None==k for k in key` (see the link in the comment) which is not really compatible with numpy arrays.  Maybe `key` shouldn't be a `numpy` array though but couldn't tell? Happy to update.

I don't see any indexing tests that have a numpy array in https://github.com/finch-tensor/finch-tensor-python/blob/main/tests/test_indexing.py though...but also array-api compat would say boolean indexes are good https://data-apis.org/array-api/latest/API_specification/indexing.html#boolean-array-indexing

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -244,8 +244,9 @@ class Tensor(_Display, SparseArray):
     def __getitem__(self, key):
         if not isinstance(key, tuple):
             key = (key,)
-
-        if None in key:
+        # if key is a numpy array or similar, == does not provide a boolean
+        # see: https://docs.python.org/3/reference/expressions.html#membership-test-operations
+        if any(k is None for k in key):
             # lazy indexing mode
             key = _process_lazy_indexing(key)
         else:


hameerabbasi:
A regression test would be appreciated here!

willow-ahrens:
The text of this issue discusses boolean indexing. However, is this issue really about `A[None]`? Is a reproducer just `A[1, None, 3]`?

ilan-gold:
@willow-ahrens I think there are two things here that I was conflating.

1. Checking `None in key` doesn't work for any numpy array indexers but that is fixed now that the incoming `key` is cast to a `tuple`: https://github.com/finch-tensor/finch-tensor-python/blob/911e750fefc88f922706406b86850faae376fe6e/src/finch/tensor.py#L256-L257 Also the check appears to have disappeared
2. Separately, I discovered the lack of boolean-array indexing support, which is in the array-api (i.e., what I posted with integer indexing doesn't have to be supported).  So that is a separate matter, that point 1., but the reason that I posted this issue in the first place.  So having boolean indexing would make me happy :) I will close this issue though in the meantime.

PR #91, opened 2025-02-28
fix julia package versioning, bump finch/juliapkg

willow-ahrens:
This follows the docs for pyjuliapkg a little better. Note the discussion in https://github.com/JuliaPy/pyjuliapkg/pull/45 about why the old behavior was not thread safe. I believe this should work better. I think this should fix https://github.com/pydata/sparse/pull/767 because of the aforementioned fix to pyjuliapkg. This would remove support for envvars that dictate the version of Finch to use, in favor of modifying the juliapkg.json file. I don't think this is that much more or less convenient than the old solution, but it is correct with how the maintainer of pyjuliapkg expected the system to be used.

### DIFF
--- a/src/finch/julia.py
+++ b/src/finch/julia.py
@@ -7,4 +7,7 @@ import juliacall as jc  # noqa
 from juliacall import Main as jl  # noqa
 
 jl.seval("using Finch")
+jl.seval("using HDF5")
+jl.seval("using NPZ")
+jl.seval("using TensorMarket")
 jl.seval("using Random")
--- a/src/finch/julia.py
+++ b/src/finch/julia.py
@@ -1,36 +1,9 @@
 import os
 
-import juliapkg
-
-
-def add_package(name: str, hash: str, version: str) -> None:
-    deps = juliapkg.deps.load_cur_deps()
-    if deps.get("packages", {}).get(name, {}).get("version", None) != version:
-        juliapkg.add(name, hash, version=version)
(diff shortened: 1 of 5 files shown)


mtsokol (approved):

src/finch/julia.py
```diff
@@ -1,37 +1,13 @@
...
-    juliapkg.add(_FINCH_NAME, _FINCH_HASH, url=_FINCH_REPO_URL, dev=True)
-else:
-    add_package(_FINCH_NAME, _FINCH_HASH, _FINCH_VERSION)
-
+#To change the version of Finch used, see the documentation for pyjuliapkg here: https://github.com/JuliaPy/pyjuliapkg
+#Use pyjuliapkg to modify the `juliapkg.json` file in the root of this repo.
+#An example development json is found in `juliapkg_dev.json`
 import juliacall as jc  # noqa
-
-juliapkg.resolve()
```
mtsokol, line 33: @willow-ahrens For the pydata/sparse `notebook` job that is failing with missing dependencies, maybe we should we have `juliapkg.resolve()` here? https://github.com/pydata/sparse/actions/runs/13648682369

willow-ahrens: Yes. It's strange that tests pass here but fail there. It may be that resolve is only called on the first instance of loading juliacall, so if pydatasparse imports juliacall and then imports finch later, resolve doesn't get called with the new deps.

PR #92, opened 2025-03-04
Reinstate `juliapkg.resolve` call

mtsokol:


willow-ahrens (approved):
This looks good to me, let's try it!

willow-ahrens:
we should merge right?

willow-ahrens:
do we need to merge to try this out?

mtsokol:
Yes!

PR #93, opened 2025-03-04
Move `juliapkg.json` file to `src/finch`

mtsokol:
@hameerabbasi After checking https://github.com/pydata/sparse/pull/846 it looks like `juliapkg.json` isn't present in the wheel, causing missing Julia dependencies. Let's try by adding the file to the `include`.


pyproject.toml
```diff
@@ -1,10 +1,13 @@
...
 name = "finch-tensor"
-version = "0.2.6"
+version = "0.2.7"
 description = ""
 authors = ["Willow Ahrens <willow.marie.ahrens@gmail.com>"]
 readme = "README.md"
-packages = [{include = "finch", from = "src"}]
+packages = [
+    { include = "finch", from = "src" },
+    { include = "juliapkg.json" },
```
hameerabbasi: IIUC, `packages` are pure-Python packages. I think what you want is [`tool.poetry.include`](https://python-poetry.org/docs/pyproject#exclude-and-include)

hameerabbasi: (You also probably additionally need `format = ["sdist", "wheel"]`).

hameerabbasi (approved):
Let's do a rebase-merge, as the commit message is now incorrect. The content of the PR LGTM

pyproject.toml
mtsokol: Done! I included it as a `finch_julia_deps/juliapkg.json` because according to the docs you linked:
> When a wheel is installed, its includes are unpacked straight into the site-packages directory. Pay attention to include top level files and directories with common names like CHANGELOG.md, LICENSE, tests or docs only in sdists and not in wheels.

mtsokol:
I still had to fix a typo, so I updated the commit message already.

hameerabbasi:
I guess now we also need to move the JSON and update all references to it. Or somehow specify a source or target.

hameerabbasi:
Thanks, @mtsokol!

PR #94, opened 2025-03-11
ENH: `linalg` module, `mT` property, and multiple outputs from `@compiled`

mtsokol:
Hi @hameerabbasi,

This PR introduces improvements that my `pydata/sparse` use-case requires:
- Array API compatible `linalg` module with the first function, `vector_norm`.
- Array API `mT` array class property.
- `@compiled` decorator accepting multiple outputs.

### DIFF
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -3,19 +3,24 @@ from functools import wraps
 
 from .julia import jl
 from .typing import JuliaObj
-from typing import Any, Iterable, Iterator, Callable
+from typing import Any, Iterator, Callable
 from .tensor import Tensor
 
+IterObj = tuple | list | dict | Any
 
-def _recurse(x: Iterable | Any, /, *, f: Callable[[Any], Any]) -> Iterable | Any:
+
+def _recurse(x: IterObj, /, *, f: Callable[[Any], Any]) -> IterObj:
     if isinstance(x, tuple | list):
         return type(x)(_recurse(xi, f=f) for xi in x)
     if isinstance(x, dict):
-        return {k: _recurse(v, f=f) for k, v in x}
+        ret = {k: _recurse(v, f=f) for k, v in x}
+        if type(x) is not dict:
+            ret = type(x)(ret)
+        return ret
     return f(x)
 
 
-def _recurse_iter(x: Iterable | Any, /) -> Iterator[Any]:
+def _recurse_iter(x: IterObj, /) -> Iterator[Any]:
     if isinstance(x, tuple | list):
         for xi in x:
             yield from _recurse_iter(xi)
@@ -27,6 +32,45 @@ def _recurse_iter(x: Iterable | Any, /) -> Iterator[Any]:
 
 def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
     return x if not isinstance(x, Tensor) else lazy(x)
+
+
+def _recurse_iter_return(x: IterObj, /, *, compute_kwargs: dict[str, Any]) -> IterObj:
+    current = 0
+
+    def recursor(_: Any) -> tuple[int, Any]:
+        nonlocal current
+        ret = current
+        current += 1
+        return ret
+
+    # Make a recursive iterator of indices.
+    idx_obj = _recurse(x, f=recursor)
+    jl_computed = []
+    py_computed = []
+
+    # Collect lazy tensors; use placeholder
+    _placeholder = object()
+    for x in _recurse_iter(x):
+        if isinstance(x, Tensor) and not x.is_computed():
+            jl_computed.append(x._obj)
+            py_computed.append(_placeholder)
+        else:
+            py_computed.append(x)
+
+    if len(jl_computed) == 1:
+        # This doesn't return an iterable of arrays -- only a single array
+        jl_computed = [jl.Finch.compute(*jl_computed, **compute_kwargs)]
+    else:
+        jl_computed = jl.Finch.compute(*jl_computed, **compute_kwargs)
+
+    # Replace placeholders with computed tensors.
+    jl_computed_iter = iter(jl_computed)
+    for i in range(len(py_computed)):
+        if py_computed[i] is _placeholder:
+            py_computed[i] = Tensor(next(jl_computed_iter))
+
+    # Replace recursive indices by actual computed objects
+    return _recurse(idx_obj, f=lambda idx: py_computed[idx])
 
 
 def compiled(opt=None, *, force_materialization=False, tag: int | None = None):
@@ -51,13 +95,7 @@ def compiled(opt=None, *, force_materialization=False, tag: int | None = None):
             if tag is not None:
                 compute_kwargs["tag"] = tag
 
-            if isinstance(result, Iterable):
-                computed = jl.Finch.compute(
-                    *map(lambda x: x._obj, result), **compute_kwargs
-                )
-                return tuple(Tensor(c) for c in computed)
-
-            return Tensor(jl.Finch.compute(result._obj, **compute_kwargs))
+            return _recurse_iter_return(result, compute_kwargs=compute_kwargs)
 
         return wrapper_func
 
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -50,6 +50,13 @@ def compiled(opt=None, *, force_materialization=False, tag: int | None = None):
             )
             if tag is not None:
                 compute_kwargs["tag"] = tag
+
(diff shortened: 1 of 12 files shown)


src/finch/tensor.py
```diff
@@ -295,6 +295,13 @@ def _is_dense(self) -> bool:
     def _order(self) -> tuple[int, ...]:
         return jl.typeof(self._obj).parameters[1]
 
+    @property
+    def mT(self) -> "Tensor":
+        axes = list(range(self.ndim))
+        axes[-2:] = axes[:-3:-1]
```
```suggestion
        axes[-2], axes[-1] = axes[-1], axes[-2]
```

src/finch/linalg/_linalg.py
```diff
@@ -0,0 +1,22 @@
...
+) -> Tensor:
+    if axis is not None:
+        raise ValueError(
+            "At the moment only `None` (vector norm of a flattened array) "
+            "is supported. Got: {axis}."
+        )
+
+    result = Tensor(jl.Finch.norm(x._obj, ord))
+    if keepdims:
+        result = result.__getitem__(tuple(None for _ in range(x.ndim)))
```
hameerabbasi:
```suggestion
        result = result[tuple(None for _ in range(x.ndim))]
```

src/finch/compiled.py
```diff
@@ -50,6 +50,13 @@ def wrapper_func(*args, **kwargs):
             )
             if tag is not None:
                 compute_kwargs["tag"] = tag
+
+            if isinstance(result, Iterable):
+                computed = jl.Finch.compute(
+                    *map(lambda x: x._obj, result), **compute_kwargs
+                )
+                return tuple(Tensor(c) for c in computed)
+
```
hameerabbasi: We can remove all of this and replace the line below by:

```python
return _recurse(result, lambda x: Tensor(jl.Finch.compute(x._obj, **compute_kwargs)))
```

hameerabbasi (changes requested):
Suggested a few simplifications to the code. The overall logic of the PR LGTM.

src/finch/tensor.py
hameerabbasi: (This is more code but clearer to parse for the uninitiated)

src/finch/linalg/_linalg.py
```diff
@@ -0,0 +1,22 @@
...
+
+def vector_norm(
+    x: Tensor,
+    /,
+    *,
+    axis: int | tuple[int, ...] | None = None,
+    keepdims: bool = False,
+    ord: int | float = 2,
+) -> Tensor:
+    if axis is not None:
```
hameerabbasi, line 15: We can probably normalize the axis, and accept only `tuple(range(ndim))` here.

hameerabbasi:
A test for multiple-return would also be appreciated.

src/finch/compiled.py
mtsokol: I disagree here - the point is to call `compute` once with all lazy tensors that come out of the same compiled function at the same time.
In your line we're calling `compute` multiple times. One time for each output.

src/finch/linalg/_linalg.py
mtsokol: That's right - done!

src/finch/linalg/_linalg.py
mtsokol: Done!

src/finch/tensor.py

mtsokol:
> A test for multiple-return would also be appreciated.

Done!

src/finch/compiled.py
hameerabbasi: Hmm. I'll push to your branch with a fix for both issues. 😄

willow-ahrens:
I'm surprised that `@compiled` didn't already accept multiple outputs! The interface you've implemented is really intuitive. Thanks Mateusz.

src/finch/compiled.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/94/commits/28cf1b088a8d50787fde6288655bc851fb3b1066

hameerabbasi (approved):

src/finch/compiled.py
```diff
@@ -29,6 +34,45 @@ def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
     return x if not isinstance(x, Tensor) else lazy(x)
 
 
+def _recurse_iter_return(x: IterObj, /, *, compute_kwargs: dict[str, Any]) -> IterObj:
```
mtsokol: I would rather rename it to `_recurse_iter_compute` because we're computing here actual tensors. I don't think that `return` is informative in the function's name.

src/finch/compiled.py
```diff
@@ -29,6 +34,45 @@ def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
...
+    current = 0
+
+    def recursor(_: Any) -> tuple[int, Any]:
+        nonlocal current
+        ret = current
+        current += 1
+        return ret
```
mtsokol: Just my preference, but I consider `global` and `nonlocal` rather obfuscating, and I think return type is incorrect here.
I think `current` and `recursor` names don't reflect the actual purpose they serve - we just want to record/visit all input args (regardless how nested they are). WDYT we have a simple ArgRecorder (or any better name) dataclass like this:
```python
@dataclass
class ArgRecorder:
    acc: int = 0
    def count(self, _) -> int:
        self.acc += 1
        return self.acc
idx_obj = _recurse(x, f=ArgRecorder().count)
```
This way we won't have the nonlocal `current = 0`.

src/finch/compiled.py
```diff
@@ -29,6 +34,45 @@ def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
...
+    py_computed = []
+
+    # Collect lazy tensors; use placeholder
+    _placeholder = object()
+    for x in _recurse_iter(x):
+        if isinstance(x, Tensor) and not x.is_computed():
+            jl_computed.append(x._obj)
+            py_computed.append(_placeholder)
+        else:
+            py_computed.append(x)
```
mtsokol: This path is used when one of the arguments is e.g. just an `int`, right? (non-Tensor things)

src/finch/compiled.py
```diff
@@ -29,6 +34,45 @@ def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
...
+
+    if len(jl_computed) == 1:
+        # This doesn't return an iterable of arrays -- only a single array
+        jl_computed = [jl.Finch.compute(*jl_computed, **compute_kwargs)]
+    else:
+        jl_computed = jl.Finch.compute(*jl_computed, **compute_kwargs)
```
mtsokol: Nitpick: Just my preference, works either way, but how about:
```python
jl_len = len(jl_computed)
jl_computed = jl.Finch.compute(*jl_computed, **compute_kwargs)
if jl_len == 1:
    jl_computed = [jl_computed]
```

mtsokol (commented):
@hameerabbasi Thanks for improving the nesting returns logic!
Left last comments, and maybe one last thing is a test which covers:
```python
@compiled()
def foo_bar(a: Tensor, b: Tensor, c: int):
    c = a + b
    d = c ** 2
    return (a, c, (b, "text"), {"key1": 12, "key2": d})
```

src/finch/compiled.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/94/commits/61455d58946d2f5fe8858c39eb9e14b55b941adc.

src/finch/compiled.py
hameerabbasi: That's much simpler indeed! Done in https://github.com/finch-tensor/finch-tensor-python/pull/94/commits/61455d58946d2f5fe8858c39eb9e14b55b941adc.

src/finch/compiled.py
hameerabbasi: Good suggestion! Done in https://github.com/finch-tensor/finch-tensor-python/pull/94/commits/61455d58946d2f5fe8858c39eb9e14b55b941adc.

hameerabbasi:
The test caught some bugs! Thanks for the suggestions, @mtsokol.

src/finch/compiled.py
hameerabbasi: Yes, I wasn't 100% sure if we could call `jl.Finch.compute` on these things. Anyway, I felt it was better not to.

PR #96, opened 2025-03-21
0.2.9 release

mtsokol:
Hi @hameerabbasi,

Here's a new release with latest Finch.jl version.

Should we keep `pyproject.toml` and `pixi.toml` aligned? I think keeping all dependencies' versions the same avoids confusion.
Do we need both `pyproject.toml` and `pixi.toml`? Can pixi use `pyproject`?
Do we need `poetry.lock` in the repo?

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -4,7 +4,7 @@ channels = ["conda-forge"]
 description = "Add a short description here"
 name = "finch-tensor"
 platforms = ["osx-arm64"]
-version = "0.1.0"
+version = "0.2.9"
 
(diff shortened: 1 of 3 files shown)


hameerabbasi:
> Do we need both `pyproject.toml` and `pixi.toml`? Can pixi use `pyproject`?

`pixi.toml` doesn't work with Poetry as it doesn't use `project.dependencies`. Thanks for keeping everything aligned!

hameerabbasi (approved):
LGTM! Thanks.

PR #97, opened 2025-03-28
`@` via `sum`

hameerabbasi:
- **Lazy indexing of `Ellipsis`, `@` via `sum` and typing improvements.**
- **`pre-commit autoupdate`**

### DIFF
--- a/src/finch/compiled.py
+++ b/src/finch/compiled.py
@@ -1,11 +1,15 @@
+from __future__ import annotations
+
 from abc import abstractmethod
 from functools import wraps
 from dataclasses import dataclass
 
 from .julia import jl
 from .typing import JuliaObj
-from typing import Any, Iterator, Callable
-from .tensor import Tensor
+from typing import Any, Iterator, Callable, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .tensor import Tensor
 
 IterObj = tuple | list | dict | Any
 
@@ -34,6 +38,8 @@ def _recurse_iter(x: IterObj, /) -> Iterator[Any]:
 
 
 def _to_lazy_tensor(x: Tensor | Any, /) -> Tensor | Any:
+    from .tensor import Tensor
+
     return x if not isinstance(x, Tensor) else lazy(x)
 
 
@@ -48,6 +54,8 @@ class _ArgumentIndexer:
 
 
 def _recurse_iter_compute(x: IterObj, /, *, compute_kwargs: dict[str, Any]) -> IterObj:
+    from .tensor import Tensor
+
     # Make a recursive iterator of indices.
     idx_obj = _recurse(x, f=_ArgumentIndexer().index)
     jl_computed = []
@@ -82,6 +90,8 @@ def compiled(opt=None, *, force_materialization=False, tag: int | None = None):
     def inner(func):
         @wraps(func)
         def wrapper_func(*args, **kwargs):
+            from .tensor import Tensor
+
             args = tuple(args)
             kwargs = dict(kwargs)
             compute_at_end = force_materialization or all(
@@ -131,6 +141,8 @@ def set_optimizer(opt: AbstractScheduler) -> None:
 
 
 def lazy(tensor: Tensor) -> Tensor:
+    from .tensor import Tensor
+
     if tensor.is_computed():
         return Tensor(jl.Finch.LazyTensor(tensor._obj))
     return tensor
@@ -139,6 +151,8 @@ def lazy(tensor: Tensor) -> Tensor:
 def compute(
     tensor: Tensor, *, opt: AbstractScheduler | None = None, tag: int = -1
 ) -> Tensor:
+    from .tensor import Tensor
+
     if not tensor.is_computed():
         if opt is None:
             return Tensor(jl.Finch.compute(tensor._obj, tag=tag))
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -168,8 +168,14 @@ class Tensor(_Display, SparseArray):
                 f"`{self.ndim=}`, `{other.ndim=}`. Both must be greater than `0`."
             )
 
-        if self.ndim == 1 or other.ndim == 1:
-            return finch.sum(self[..., :] * other[..., :], axis=-1)
+        if self.ndim == 1 and other.ndim == 1:
+            return finch.sum(self * other, axis=-1)
+
+        if self.ndim == 1:
+            return finch.sum(self * other[..., :, :].mT, axis=-1)
+
+        if other.ndim == 1:
+            return finch.sum(self * other, axis=-1)
 
         return finch.sum(self[..., :, None, :] * other.mT[..., None, :, :], axis=-1)
 
@@ -255,7 +261,7 @@ class Tensor(_Display, SparseArray):
         if not isinstance(key, tuple):
             key = (key,)
 
-        if None in key:
+        if not self.is_computed():
             # lazy indexing mode
             key = _process_lazy_indexing(key, self.ndim)
         else:
@@ -1273,7 +1279,7 @@ def _expand_ellipsis(key: tuple, shape: tuple[int, ...]) -> tuple:
     key_without_ellipsis = ()
     # first we need to find the ellipsis and confirm it's the only one
     for pos, idx in enumerate(key):
-        if idx == Ellipsis:
+        if idx is Ellipsis:
             if ellipsis_pos is None:
                 ellipsis_pos = pos
             else:
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -161,23 +161,18 @@ class Tensor(_Display, SparseArray):
 
     @compiled()
     def __matmul__(self, other: Tensor) -> Tensor:
-        import finch
-
         if self.ndim == 0 or other.ndim == 0:
             raise ValueError(
                 f"`{self.ndim=}`, `{other.ndim=}`. Both must be greater than `0`."
             )
 
-        if self.ndim == 1 and other.ndim == 1:
-            return finch.sum(self * other, axis=-1)
+        if other.ndim == 1:
+            return sum(self * other, axis=-1)
 
         if self.ndim == 1:
-            return finch.sum(self * other[..., :, :].mT, axis=-1)
+            return sum(self * other.mT, axis=-1)
 
-        if other.ndim == 1:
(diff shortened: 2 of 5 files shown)


hameerabbasi:
@willow-ahrens I seem to have found a bug in the default scheduler. The test that I added for 3D stacked matrix multiplies passes for the galley scheduler but fails for the default scheduler. You can run the test for both schedulers with `pytest -k matmul`.

hameerabbasi:
Hmm, I'm not sure what the difference is, but I'm able to reproduce it locally on macOS, Python 3.12. I see some environments on CI aren't able to reproduce the bug. Let me know if you'd like to pair program to pin down the issue.

src/finch/tensor.py
```diff
@@ -245,12 +261,12 @@ def __getitem__(self, key):
...
 
-        if None in key:
+        if not self.is_computed():
             # lazy indexing mode
-            key = _process_lazy_indexing(key)
+            key = _process_lazy_indexing(key, self.ndim)
         else:
             # standard indexing mode
-            key = _expand_ellipsis(key, self.shape)
+            key = _expand_ellipsis(key, self.ndim)
```
mtsokol: This isn't correct. The signature is `_expand_ellipsis(key: tuple, shape: tuple[int, ...])` so you need to either reverse it or change `_expand_ellipsis` implementation.

src/finch/tensor.py
```diff
@@ -245,12 +261,12 @@ def __getitem__(self, key):
         if not isinstance(key, tuple):
             key = (key,)
 
-        if None in key:
+        if not self.is_computed():
```
mtsokol, line 259: I think we still want to support lazy indexing in eager mode, why we need this change?

src/finch/tensor.py
```diff
@@ -156,12 +159,25 @@ def __mod__(self, other):
     def __pow__(self, other):
         return self._elemwise_op("^", other)
 
-    def __matmul__(self, other: "Tensor") -> "Tensor":
-        if self.ndim < 2 or other.ndim < 2:
-            raise NotImplementedError(
-                "`@` is not implemented for arrays with less than two dimensions."
+    @compiled()
+    def __matmul__(self, other: Tensor) -> Tensor:
```
mtsokol, line 163: Can we do it the other way around?
Instead of using `complied()` here, can we import `from .compiled import lazy, compute` inside `def __matmul__` and wrap `self` and `other`? This way we can revert `compiled.py` file changes completely.

mtsokol (changes requested):

src/finch/tensor.py
mtsokol: This should solve CI errors.

src/finch/tensor.py
hameerabbasi: There are a number of issues with this approach:
- ~~It won't get past the recursive import since we're still doing `from .compiled import ...`~~ I just realised you suggested doing it inside the function. There's still the second issue.
- It will directly compute the matmul, even if it is part of a lazy evaluation, which is what we'd like to avoid.

src/finch/tensor.py
hameerabbasi: I've added support for `None` in eager indexing in this commit (after this comment): https://github.com/finch-tensor/finch-tensor-python/pull/97/commits/e4bd64c35d1c6d57b169270e633d65ceb64907aa

hameerabbasi:
> Hmm, I'm not sure what the difference is, but I'm able to reproduce it locally on macOS, Python 3.12. I see some environments on CI aren't able to reproduce the bug. Let me know if you'd like to pair program to pin down the issue.

Now we're back to this error once more, the indexing bugs have been fixed.

src/finch/tensor.py
mtsokol: Sure! Thanks for an explanation!

mtsokol:
@hameerabbasi Once CI is green I still want to run `@`-related benchmarks before approving to make sure we don't deteriorate the speedup of existing examples.

mtsokol:
I reproduced CI issue here: https://github.com/finch-tensor/Finch.jl/issues/720

hameerabbasi:
> I reproduced CI issue here: [finch-tensor/Finch.jl#720](https://github.com/finch-tensor/Finch.jl/issues/720)

The fix is in https://github.com/finch-tensor/finch-tensor-python/pull/97/commits/c95972abb7b365f9334f4f24bb083127d090d824.

hameerabbasi:
On my machine, 

`examples/matmul_example.py` in `sparse` returns (with 100 iterations instead of 3):
This branch: `0.01588613271713257 s`
`main`: `0.016171562671661376 s`

src/finch/tensor.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
mtsokol, line 1: What does it do?

tests/test_ops.py
```diff
@@ -288,21 +294,34 @@ def test_tensordot(arr3d, storage, opt):
...
+    expected = a @ b
     actual = A_finch @ B_finch
-    expected = arr2d @ arr2d.T
-    assert_equal(actual.todense(), expected)
 
-    actual = A_finch @ C_finch
     assert_equal(actual.todense(), expected)
 
-    with pytest.raises(Exception, match="DimensionMismatch"):
-        A_finch @ D_finch
```
mtsokol, line 305: It's still a branch to test - a case where there's a dimensions mismatch

src/finch/compiled.py
```diff
@@ -48,6 +54,8 @@ def index(self, _) -> int:
 
 
 def _recurse_iter_compute(x: IterObj, /, *, compute_kwargs: dict[str, Any]) -> IterObj:
+    from .tensor import Tensor
```
willow-ahrens, line 57: I'm somewhat confused why these have been added. did this not work before?

src/finch/tensor.py
```diff
@@ -156,12 +159,20 @@ def __mod__(self, other):
...
             )
-        return tensordot(self, other, axes=((-1,), (-2,)))
+
+        if other.ndim == 1:
+            return sum(self * other, axis=-1)
+
+        if self.ndim == 1:
+            return sum(self * other.mT, axis=-1)
+
+        return sum(self[..., :, None, :] * other.mT[..., None, :, :], axis=-1)
```
willow-ahrens, line 175: Is this all that was needed? That's quite clever!

tests/test_ops.py
```diff
@@ -288,21 +294,34 @@ def test_tensordot(arr3d, storage, opt):
     assert_equal(actual.todense(), expected)
 
 
-def test_matmul(arr2d, arr3d, opt):
-    A_finch = finch.Tensor(arr2d)
-    B_finch = finch.Tensor(arr2d.T)
-    C_finch = finch.permute_dims(A_finch, (1, 0))
-    D_finch = finch.Tensor(arr3d)
+@pytest.mark.parametrize(
```
willow-ahrens, line 298: Just to double check, these new tests are a superset of the old ones, right?

tests/test_ops.py
willow-ahrens: yes, let's leave these error tests in.

src/finch/tensor.py
```diff
@@ -380,19 +391,23 @@ def todense(self) -> np.ndarray:
             dense_tensor = dense_tensor.lvl
 
         result = np.asarray(jl.reshape(dense_tensor.val, shape))
-        return result.transpose(self.get_inv_order()) if self._is_dense else result
+        return (
+            result.transpose(tuple(o - 1 for o in self._order))
```
willow-ahrens: This is the only concerning change to me. It can be very difficult to sufficiently test the difference between get_inv_order and o - 1 because we must use dimension != 2. I suppose your fix here is saying that when we densify, we aren't really changing the order at all so there's no reason to inv_order?

willow-ahrens (commented):
Things mostly look good to me! I'm not well-versed in python typing or the annotations feature, and my comments are only clarifications.

src/finch/compiled.py
mtsokol: Here it's required to avoid a cyclic import. As in this PR `def __matmul__` method in `Tensor` is now `@compiled`, `tensor.py` imports `compiled.py`. `compiled.py` in general imports `tensor.py` so we need to import it lazily now in relevant functions.

src/finch/tensor.py
mtsokol: With stacked matmul (implemented with lazy indexing) it seems that `get_inv_order` was wrongly used in the first place and `todense()` was broken. Previously we had only 2dim matmul by tensordot.

src/finch/tensor.py
hameerabbasi: Yes! The lines above generalise this function to a vector-vector, matrix-vector and vector-matrix product, and this one is for a matrix-matrix product.

tests/test_ops.py
hameerabbasi: Correct, other than the one you point out below for a dimension mismatch. I'll add a separate test for that one.

src/finch/tensor.py
hameerabbasi: See also #98 for a bug-fix and regression test.

src/finch/tensor.py
hameerabbasi: This changes all annotations to be interpreted like strings. This has a number of ergonomics advantages. For example, in the following two situations we don't need string annotations and can use normal type annotations.

```python
from __future__ import annotations

if TYPE_CHECKING:
    from ... import MyClass

def f() -> MyClass:
    ...

class MyClass2:
    def f2(self) -> MyClass2:
        ...
```

tests/test_ops.py
hameerabbasi: Done in https://github.com/finch-tensor/finch-tensor-python/pull/97/commits/0719f502f5b9239d6918006ab541f9effc7c78f7.

mtsokol:
I think there's an error in the logic that expands `...` in the index to `:` symbols.

hameerabbasi:
> I think there's an error in the logic that expands `...` in the index to `:` symbols.

Fixed in https://github.com/finch-tensor/finch-tensor-python/pull/97/commits/8ff3876f69b137a0e1c58a70c8867bebb7c0f587.

mtsokol (approved):
Thanks @hameerabbasi!

PR #98, opened 2025-03-31
BUG: Fix order handling in `todense()`

mtsokol:
Hi @hameerabbasi,

I separated `todense()` fix here and co-authored you in the commit if it's Ok.

### DIFF
--- a/src/finch/tensor.py
+++ b/src/finch/tensor.py
@@ -380,7 +380,7 @@ class Tensor(_Display, SparseArray):
             dense_tensor = dense_tensor.lvl
 
         result = np.asarray(jl.reshape(dense_tensor.val, shape))
-        return result.transpose(self.get_inv_order()) if self._is_dense else result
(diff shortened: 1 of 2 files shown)


tests/test_ops.py
hameerabbasi, line 1: Can we revert this file to avoid any conflicts with the matmul PR?

hameerabbasi (changes requested):
Looks mostly good. Let's add a regression test.  Something using `.permute_dims((2, 0, 1))` should be enough.

tests/test_ops.py
mtsokol: Done!

hameerabbasi:
A regression test would be nice.

mtsokol:
> A regression test would be nice.

I extended `test_permute_dims` so now it covers cases which fail for `main` branch.

hameerabbasi (approved):
Thanks, @mtsokol!

PR #121, opened 2025-04-25
CI: Add `array-api-tests` CI job and upgrade Array API version

mtsokol:
Hi @willow-ahrens,

Here's a copy of `array-api-tests` job from `pydata/sparse` but with skips only to speed it up.


mtsokol:
@hameerabbasi This PR adds `array-api-tests` to `finch-tensor-python` for better feedback loop (if anything breaks we know before the release of `finch-tensor`, instead of `pydata/sparse` CI).

I reverted `reshape` implementation to the temporary one, as we need an issue to be solved.

Here we have all failing tests in skips instead of skips+xfails - it's meant to be just a quick sane check, the actual xfails are tested in `pydata/sparse`.

I also upgraded Array API to `2024.12` and added all necessary skips for new functions/changes. From here we can continue addressing them separately.
I already added `__array_namespace_info__` (taken mostly from NumPy)

mtsokol:
@hameerabbasi Would you have time to review it?

src/finch/tensor.py
```diff
@@ -717,12 +717,13 @@ def reshape(
...
-    obj = jl.swizzle(obj, *tuple(reversed(range(1, jl.ndims(obj) + 1))))
-    return Tensor(obj)
-
+    # TODO: https://github.com/finch-tensor/Finch.jl/issues/743
+    #       Revert to `jl.reshape` implementation once aforementioned
+    #       issue is solved.
+    warnings.warn("`reshape` densified the input tensor.", PerformanceWarning)
+    arr = x.todense()
+    arr = arr.reshape(shape)
+    return Tensor(arr)
```
hameerabbasi, line 726: I would revert this and leave the job as failing for now.

hameerabbasi (changes requested):
LGTM besides the one change. I'd still advocate for strict xfails over skips, but that may just be me.

mtsokol:
> LGTM besides the one change. I'd still advocate for strict xfails over skips, but that may just be me.

Thanks! We still have complete xfails in `pydata/sparse`, here it's more for the purpose of quicker feedback loop.

src/finch/tensor.py
mtsokol: The problem here is that `reshape` function is used in multitude of Array API tests, for the input setup. It's not only `test_reshape` but a great deal of Array API tests (otherwise I'd also left it failing but now `array-api-tests` job in `pydata/sparse` is unusable).

hameerabbasi (approved):

hameerabbasi:
Thanks, @mtsokol!

PR #123, opened 2025-09-04
get ready to update to finch-tensor and publish

willow-ahrens:
Update the publish yml with the new pyproject.toml layout, also bump the version and rename finch-tensor-python to finch-tensor

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,7 +15,7 @@ jobs:
         run: pip install toml
       - name: Get version
         run: |
-            VERSION=$(python -c 'import toml; print("v" + toml.load("pyproject.toml")["tool"]["poetry"]["version"])')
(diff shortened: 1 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         run: pip install toml
       - name: Get version
         run: |
-            VERSION=$(python -c 'import toml; print("v" + toml.load("pyproject.toml")["tool"]["poetry"]["version"])')
+            VERSION=$(python -c 'import toml; print("v" + toml.load("pyproject.toml")["project"]["version"])')
```
mtsokol, line 18: Confusing for sure, as in `pyproject.toml` the version is under `[tool.poetry]` section.

mtsokol (approved):

PR #125, opened 2025-10-14
Fix publish

willow-ahrens:
Adds pre-commit linting
migrates poetry to [project] rather than [tool.poetry]


mtsokol:
Looks like ruff found out that three tests had the same name and as a result only the last one was executed (and that two tests were accidentally nested in another one).

As this PR is linting only I just skipped them - we can work on enabling them separately (they've never run before).

Last thing is this HDF5 Julia build error on M1, we already constrained Julia version to 1.11 so it must be something else.

willow-ahrens:
based on https://github.com/JuliaPy/PythonCall.jl/issues/615, simplest fix is to just bump minimum python version.

PR #132, opened 2026-02-20
Reorienting finch-tensor as a Julia backend for finch tensor lite

JoelMathewC:
This PR is an overhaul of the `finch-tensor` project. The changes here reposition this package as a julia backend for the `finchlite` project. The intention behind this revamp is (1) to make `finchlite` the singular interface to interact with the finch compiler and (2) a temporary measure to expose sparse formats for projects using the `finch-tensor-lite` interface.

### The changes in this PR are as follows
1. `finchlite.FType` inherited sparse and dense formats.
2. `finchlite.Tensor` inherited `FinchJLTensor` which wraps around raw `Finch.jl` tensor objects.
3. `FinchJLCompiler`, `FinchJLGenerator`, `FinchJLLibrary`, `FinchJLKernel` which work together to translate `finchlite.finchntn` -> `@finch` macro'd julia strings.
4. We make the simplifying assumption that the input data is ordered in column major format and thus avoid doing unnecessary transposes.

For example, the following code
```
finchlite.interface.set_default_scheduler(ctx=COMPILE_JULIA)
A_finch = FinchJLTensor(jl.Finch.Tensor(jl.Dense(jl.Dense(jl.Element(0.0))), A))
B_finch = FinchJLTensor(jl.Finch.Tensor(jl.Dense(jl.Dense(jl.Element(0.0))), B))
C = finchlite.einop("C[i,j] += A[i,k] * B[k,j]", A=A_finch, B=B_finch)
```

compiles down to
```
function main(_A_147,_A_2_151,_A_3_155)
    @finch _A_3_155 .= 0.0
    @finch begin
        for _i_159 = _
            for _i_2_160 = _
                for _i_3_161 = _
                    _A_3_155[_i_159,_i_3_161] += *(_A_147[_i_159,_i_2_160],_A_2_151[_i_2_160,_i_3_161])                
                end
            end
        end
    end
    return _A_3_155
end
```

### Some project setup changes
1. The python version requirement has shifted to `requires-python = ">=3.11,<3.14"` because finchlite depends on these versions.
2. Added `finch-tensor-lite` as a dependency

### Regression
1. The changes here have rendered the previous versions of `zeros()`, `zeros_like()`, `full()`, `full_like()`, `ones()`, `ones_like()`, `reshape()`, `.asarray()` and a few other functions invalid. This means that `array-api-tests/` are not passing.


willow-ahrens (changes requested):
Let's add an ftype for the outer tensor as well.

JoelMathewC:
@willow-ahrens this is the PR to make `finch-tensor` a julia backend for `finch-tensor-lite`. The changes in this PR do not provide feature parity with the old version of `finch-tensor` specifically the array APIs. I expect that fixing those might take a considerable amount of time since we'd have to go through them case by case and account for anything we missed. However, this is at a place where GLAM can use this for codegen.

willow-ahrens:
closed in favor of #134

PR #135, opened 2026-03-30
Wma/array api 2

willow-ahrens:


tests/test_array_api.py
```diff
@@ -0,0 +1,147 @@
...
+    )
+    print(f"[array-api] user nested pytest args: {ARRAY_API_TESTS_ARGS}", flush=True)
+    result = subprocess.run(
+        [
+            sys.executable,
+            "-m",
+            "pytest",
+            *NESTED_PYTEST_ARGS,
+            f"{ARRAY_API_TESTS_DIR}/array_api_tests/",
+            "--max-examples=2",
```
mtsokol, line 135: I think in the future this should be parametrized as well - 2 examples mostly show that the function can be imported and that the simplest case passes. Maybe for finch-tensor-lite it's worth investing time to make it to e.g. 100.

tests/test_array_api.py
```diff
@@ -0,0 +1,147 @@
...
+    FORWARDED_MAIN_PYTEST_ARGS = _get_forwarded_main_pytest_args(request)
+    ARRAY_API_TESTS_ARGS = _get_user_array_api_pytest_args(request)
+    NESTED_PYTEST_ARGS = [*FORWARDED_MAIN_PYTEST_ARGS, *ARRAY_API_TESTS_ARGS]
+
+    print(f"[array-api] using dir: {ARRAY_API_TESTS_DIR}", flush=True)
+    print(f"[array-api] target rev: {ARRAY_API_TESTS_REV}", flush=True)
+
+    if not os.path.isdir(ARRAY_API_TESTS_DIR):
+        print("[array-api] cloning test repo...", flush=True)
+        subprocess.run(
```
mtsokol, line 68: If we run this test locally twice, won't it complain that it's already cloned?

mtsokol (approved):
LGTM - these 5 tests on windows - let's skip them for now.

tests/test_array_api.py
mtsokol: Ah, I see there's a check before it.

